PR #8, opened 2014-03-02
make tests run in current directory rather than sub

### LINKED ISSUE #7: Move code out of subdirectory
In order to match the other language directories, the exercises need to be in the root of the project.

Right now moving would break the build, because the tests try to run against ALL the directories in the relevant directory, and it doesn't exclude `.git` -- so the Makefile will need to be updated.

wilmoore:
fixes #7.


petemcfarlane:
[Move oAuth route](https://trello.com/c/oj8ppfMe/365-move-oauth-route)

PR #37, opened 2015-01-08
Roman: Changed from Class to Function. Used another built in function.

Dog:
Took out the while loop and used str_repeat instead.


kytrinyx:
Nice, thanks!

PR #45, opened 2015-03-18
Updated test to check for different length DNA strands

petemcfarlane:
I thought this would be a useful thing to check for as it is mentioned in the X-common implementation notes.


kytrinyx:
Nice, thanks!

PR #46, opened 2015-03-19
Fixed test, swapped 'expected' with 'actual' parameters

petemcfarlane:


kytrinyx:
I find myself swapping these in almost every language that I write, and it doesn't help that every once in a while the test framework expects them the other way around. Thanks :)

PR #49, opened 2015-03-25
Skip all but the first test.

dkinzer:
By skipping the all but the first tests we

(1) Do not overwhelm the user with a bunch of failing tests.
(2) Are more in line with TDD development.

Setting up the tests as proposed would be in line with the way that the ruby
tests have been set up.

### DIFF
--- a/bowling/bowling_test.php
+++ b/bowling/bowling_test.php
@@ -25,6 +25,7 @@ class GameTest extends \PHPUnit_Framework_TestCase
 
     public function testAllOnes()
     {
+        $this->markTestSkipped();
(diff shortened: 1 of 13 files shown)


kytrinyx:
Thanks!

dkinzer:
Hmmm.. ... good point.  phpunit seems to have a --stop-on-failure parameter which might do the trick.  But not for regular folks who wont know to run it with that parameter.  Maybe piping would work here as well.

kytrinyx:
We could put the instructions in the SETUP.md so that it's delivered with the README.

We could also put this in the website documentation.

PR #50, opened 2015-03-25
Run all tests in development.

dkinzer:
This commit assures that tests are run on Travis even if test have been marked
as skipped.

See exercism/xphp#49

PR #51, opened 2015-03-26
Fix make test not working in macosx.

dkinzer:
It turns out that sed -i in macosx works differently then on linux.
This commit uses a pipe to sed delete the skips avoiding -i
altogether.

Also, this commit adds the word "Skip" to a Gigasecond test that we do
want to be skipped.


dkinzer:
Sorry for all the pull requests.  This fixes a potential issue introduced by #50, and as a bonus it also cleans up code a little.

kytrinyx:
It's all good. Thanks for cleaning this up and getting it all sorted!

PR #53, opened 2015-04-11
Fixing mixed up expected and actual values for Gigasecond test

peteraba:
All the other exercises seem to use the asserts properly.


kytrinyx:
Nice catch, thanks.

PR #55, opened 2015-04-18
Fix the robot-name test case including the example

schorsch3000:
I've fixed the testHasName().
It let really wired strings pass the test.
Also I've fixed the example to give valid names since it was generating 5 to 7 digit numbers as name (which were passing the wired test) .

### DIFF
--- a/robot-name/example.php
+++ b/robot-name/example.php
@@ -8,7 +8,7 @@ class Robot
 
     public function __construct()
     {
-        $this->alphabet = str_split('ABCDEFGHIJKLMNOPQRSTUVWXYZ');
+        $this->alphabet = array_flip(str_split('ABCDEFGHIJKLMNOPQRSTUVWXYZ'));
     }
 
     public function getName()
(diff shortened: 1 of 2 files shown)


kytrinyx:
Good catch. I'll let the track maintainers merge this, but I think you're right about the tests being too permissive.

Dog:
Could you write your commit messages in the style <exercise name>: <what you did>? We don't currently have a section for preferred style on commits, but this format is used in other tracks. Its helped make reading the git history easier.

I also noticed that you don't seem to have your gitconfig set up for GitHub. You may want to set it up, and amend your commits so you will get credit for your work.

If you need help doing this let us know and we can help you through it. If doing so is too intimidating then we can tweak it before merging.

https://help.github.com/articles/set-up-git/#setting-up-git
https://www.atlassian.com/git/tutorials/rewriting-history/git-commit--amend

schorsch3000:
I've set up my email-address correct and edited the commit-messages according to the preferred style.

kytrinyx:
Would you run `git push -f origin master` to update the PR, please?

Dog:
Travis ran into an issue with PHP today, which caused the build to fail. I put the code back into the queue to be rebuilt, but they have a really large backlog right now.

Dog:
This was merged in cda9fa9.

Thanks!

PR #57, opened 2015-07-29
Ensure long diagonals don't count as adjacent

karptonite:
I noticed [one solution](http://exercism.io/submissions/59bd740ef4664d1eb5ac6ab9e35d8815) to the connect exercise that passed the test, but treated the board like a rectangular grid, counting "illegal" long diagonals as connections. This added test should catch those incorrect algorithms.


kytrinyx:
Nice! Great catch, thanks.

PR #59, opened 2015-11-28
Force the gigasecond exercise to use a immutable version of Datetime

TFarla:
The current solution in the example has side effects.
```php
function from(\DateTime $from)
{
    $interval = new DateInterval("PT1000000000S");
    return $from->add($interval);
}
```

The `add` method not only returns a new `DateTime` but it also mutates `$from`.
Given that I would call this function twice with the same value.

```php
$date = new DateTime('1990-01-01');
from($date); // $date is now 2021-09-09
from($date); // $date is now 2053-05-18
```

This is clearly bad practice and is easily solved by using [DateTimeImmutable](http://php.net/manual/en/class.datetimeimmutable.php).

### DIFF
--- a/gigasecond/example.php
+++ b/gigasecond/example.php
@@ -1,6 +1,6 @@
 <?php
 
-function from(\DateTime $from)
+function from(\DateTimeImmutable $from)
 {
     $interval = new DateInterval("PT1000000000S");
     return $from->add($interval);
--- a/gigasecond/example.php
+++ b/gigasecond/example.php
@@ -1,7 +1,8 @@
 <?php
 
-function from(\DateTimeImmutable $from)
+function from(\DateTime $from)
 {
     $interval = new DateInterval("PT1000000000S");
-    return $from->add($interval);
+    $date = clone $from;
+    return $date->add($interval);
 }
--- a/gigasecond/gigasecond_test.php
+++ b/gigasecond/gigasecond_test.php
@@ -8,7 +8,7 @@ class GigasecondTest extends \PHPUnit_Framework_TestCase
     public function dateSetup($date)
     {
         $UTC = new DateTimeZone("UTC");
-        $date = new DateTimeImmutable($date, $UTC);
+        $date = new DateTime($date, $UTC);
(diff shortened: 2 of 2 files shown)


kytrinyx:
Yes, good catch. The build is failing. Would you mind taking a look at it?

@Dog do you see any gotchas with this?

TFarla:
@kytrinyx It fails with the following error:
```
running tests for: bowling
This version of PHPUnit requires PHP 5.6; using the latest version of PHP is highly recommended.
make[1]: *** [test-assignment] Error 1
make: *** [test] Error 1
```
The travis build uses php 5.5 so its a compatibility issue and not an issue with the tests themselves.
A solution would be to update the `.travis.yml` to use php 5.6

Dog:
I feel it really depends on what you are going to use gigasecond for, and the intent isn't really given from the readme.


Functionally, I see nothing wrong with this:

```
$date = new DateTime('1990-01-01');
from($date); // $date is now 2021-09-09
from($date); // $date is now 2053-05-18
```

It could be argued that you don't want to modify the original date time, but thats really left up to the implementor and the specification. I definitely agree the solution has a problem, but the issue isn't inherently its functionality. I think the break down was I assumed add would return a new Datetime without touching the old one - similar to the DateTime library in Python.

The only thing that bothers me about this solution, is I don't like the assumption you will be passed an immutable date time. What I would prefer is a solution that created a new DateTime object without modifying the original - that is regardless of if its input is immutable. We could then add in a test to check the original date time wasn't modified.

kytrinyx:
I like the direction of not constraining the solution space directly, but adding a test to check for immutability.

TFarla:
@Dog The mutable behaviour looks too much like an output argument to me and that didn't sit well. I modified the example to match your requirements.

Thank you for taking the time to respond!

petemcfarlane:
This can probably be tried again on Travis by someone with permissions, and then merged.

kytrinyx:
I'll restart the build. Thanks, @petemcfarlane

PR #62, opened 2015-12-13
Add hello world problem

duffn:
Closes https://github.com/exercism/todo/issues/42.


petemcfarlane:
This build failed because at the time Travis was testing with PHP 5.5. If someone with privileges can re-run the build then I think it should pass, and is good to be merged.

PR #67, opened 2016-04-02
Update ABOUT.md

garak:


docs/ABOUT.md
```diff
@@ -0,0 +1,3 @@
+PHP is a popular general-purpose scripting language that is especially suited to web development.
+
+Fast, flexible and pragmatic, PHP powers everything from your blog to the most popular websites in the world.
```
kytrinyx: This sounds a bit market-y. Can we use some specific examples of sites that use PHP?

garak: That's just the text of php.net website.
Anyway, you can easily verify that PHP runs most of Internet websites, about 82% according to [wikipedia](https://en.wikipedia.org/wiki/PHP#Use)
This is an example of [famous website](https://facebook.com) ran by PHP :stuck_out_tongue_winking_eye:

kytrinyx: Right, let's be specific. Facebook and WordPress are both good examples. A lot of sites are built on Drupal, as well.

petemcfarlane:
Closing to clean up the PRs, this seems quite out of date. Feel free to reopen if I'm wrong!

PR #71, opened 2016-04-08
Fix issue 70, improve clock tests.

petemcfarlane:


kytrinyx:
Thanks, Pete. I can't remember the deal in the PHP track. Did we decide to remove the skips from the exercises?

petemcfarlane:
Nope, that’s my fault! I couldn’t get the makefile working so I was running the tests from the cli instead. I’ll correct that now!

kytrinyx:
Ah, let me see if I can figure out what's going on with the makefile.

kytrinyx:
Oh. Uh--the make file isn't actually running any tests at all on CI. I just fixed the same thing on the Ruby track this morning. Oops.

petemcfarlane:
Ahh good spot! I've fixed that in the last commit 976261d :+1:

kytrinyx:
That travis run looks *much* better.

Would you mind sticking the makefile fix into a different PR? If you're on you way to work or something I can cherrypick it locally.

petemcfarlane:
Yup, I've moved it into PR #72

kytrinyx:
Sweet, thanks so much!

PR #73, opened 2016-05-06
Isogram: Tests and Example

ecrmnn:
This PR adds tests and example for the Isogram exercise.

Tests taken from [x-common/isogram.json](https://github.com/exercism/x-common/blob/master/isogram.json)

### DIFF
--- a/config.json
+++ b/config.json
@@ -8,6 +8,7 @@
     "gigasecond",
     "rna-transcription",
     "raindrops",
+    "isogram",
     "difference-of-squares",
     "roman-numerals",
     "robot-name",
--- a/exercises/isogram/example.php
+++ b/exercises/isogram/example.php
@@ -1,0 +1,8 @@
+<?php
+
+function isIsogram($string)
+{
+    $string = str_replace(['-', ' '], '', mb_strtolower($string));
+    $letters = preg_split('//u', $string, -1, PREG_SPLIT_NO_EMPTY);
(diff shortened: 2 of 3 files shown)


petemcfarlane:
This would be better as a function `isIsogram($input)` rather than a class. 

I can see you've taken the test cases from the [x-common/isogram.json](https://github.com/exercism/x-common/blob/d5fe6ff52bab3ca233b9eb76985fc0ce355f187b/isogram.json) which is great, but I think the test names could be a bit clearer, for example `testAlphabet` could be `testCaseInsensitive`, `testEmilyJungSchwartzkopf` could be `testIgnoresSpaces`. That will be clearer for the person doing the exercise to figure out what you are testing, and it also serves as good documentation for someone in the future to see how the function works.

Finally, I have a more succinct solution which you can use for the example, if you want. 
```php
function isIsogram($input)
{
    $input = mb_strtolower($input);
    $input = str_replace(['-', ' '], '', $input);
    $letters = preg_split('/(.{1})/us', $input, null, PREG_SPLIT_NO_EMPTY | PREG_SPLIT_DELIM_CAPTURE);
    $uniqueLetters = array_unique($letters);
    return $letters == $uniqueLetters;
}
```
Thanks again for all your efforts!

ecrmnn:
Refactored. Should be good to go now.

kytrinyx:
Thank you!

PR #74, opened 2016-05-06
Anagram: Tests and Example

ecrmnn:
This PR adds tests and example for the Anagram exercise.

Tests taken from [x-common/anagram.json](https://github.com/exercism/x-common/blob/master/anagram.json)

### DIFF
--- a/config.json
+++ b/config.json
@@ -12,6 +12,7 @@
     "roman-numerals",
     "robot-name",
     "leap",
+    "anagram",
     "trinary",
     "bowling",
     "clock",
--- a/exercises/anagram/anagram_test.php
+++ b/exercises/anagram/anagram_test.php
@@ -1,0 +1,128 @@
+<?php
+
+require "anagram.php";
+
+class AnagramTest extends PHPUnit_Framework_TestCase
+{
+    public function testNoMatches()
+    {
+        $this->assertEquals([], detectAnagrams('diaper', ['hello', 'world', 'zombies', 'pants']));
+    }
+
+    public function testDetectsSimpleAnagram()
+    {
+        $this->markTestSkipped();
+        $this->assertEquals(['tan'], detectAnagrams('ant', ['tan', 'stand', 'at']));
+    }
+
+    public function testDoesNotDetectFalsePositives()
+    {
+        $this->markTestSkipped();
+        $this->assertEquals([], detectAnagrams('galea', ['eagle']));
+    }
+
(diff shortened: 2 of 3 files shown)


petemcfarlane:
Just my opinion again, but I don't think the anagram detector really needs to be an object 😁 
I think `detectAnagrams('master', ['stream', 'pigeon', 'maters'])` would be much cleaner. In other tracks, including [Python](https://github.com/exercism/xpython/blob/4e4f1e5b65617f6ad419d2df5a730a43ebd59eeb/exercises/anagram/anagram_test.py), they use a function but in others such as [JavaScript](https://github.com/exercism/xjavascript/blob/9cfb49cd21aba663da308e72580064e95ac941dd/exercises/anagram/anagram.spec.js) an object is used so there seems to be some split ideas.

Theres a couple of tests that I'm not sure are needed
 - `testDetectsUnicodeAnagrams` and `testEliminatesMisleadingUnicodeAnagrams` - do we need this extra level of complexity?
 - `testAnagramsMustUseAllLettersExactlyOnce2` - this seems very similar to a previous test and I think is redundant.
 - `testAcceptsStringArguments` and `testAcceptsSingleStringArgument` - Personally I think using `func_get_args()` and determining if you're given a string, array or multiple arguments is not a very nice thing to do. It adds a new level of confusion and can introduce all kinds of bugs in production code, we shouldn't be encouraging people to use this! I would suggest only accepting an `array` of candidate words instead.

ecrmnn:
As mentioned in OP, tests are identical from ``anagram.json``. Thought it was preferable to have identical tests across all language tracks, but if you think they are redundant I'll remove them.

I don't like ``func_get_args()`` at all, just added it so you could pass unlimited strings like the JS implementation. I'll remove it and only accept an array

petemcfarlane:
You're right about the `patter/tapper` test, I'm sorry. I think in this case it would be best to leave it in for consistency (but I still think it's redundant in the master anagram.json file too really 😉)

ecrmnn:
Refactored

kytrinyx:
>  I still think it's redundant in the master anagram.json file too really

Yeah, probably. I've opened an issue on the metadata repo about it.

ecrmnn:
@kytrinyx Want me to remove the patter/tapper tests?

kytrinyx:
I haven't heard back from the other maintainers on this (in the above linked issue). Let's leave it as is for now.

Issue #29, opened 2014-11-21
Classes vs Functions


Dog:
I know a few of the exercises in xpython were updated from classes with a single function to simply functions. I noticed that xphp has a few exercises where I think it'd make sense to use a function rather than a class. Two examples are hamming and leap. They accept one argument on initialization that could be the argument to the function.

No use case is given where the argument needs to be stored with its instance, or the class later being extended. I think it'd be more appropriate to define them as functions.

Does anyone else have an opinion on this?

Dog:
These are the problems I have left to go through for this issue:

- [ ] bowling
- [ ] clock
- [ ] robot-name
- [ ] wordy

petemcfarlane:
I have re-written wordy as a function #43. 

I don't think bowling, clock or robot-name make sense as functions because they have persisted states that are stored within the instance. What do you think @Dog?

kytrinyx:
Agreed -- I think bowling and robot-name should remain as they are.

petemcfarlane:
What about clock? I think it's hard to add to, subtract from or compare something that doesn't have a state.

kytrinyx:
Yeah, that's my understanding as well. Let's leave clock as is.

kytrinyx:
OK, cool. Closing this--if something comes up later we can always reopen or create a new issue.

Thanks, all!

PR #75, opened 2016-05-06
WordCount: Tests and Example

ecrmnn:
This PR adds tests and example for the WordCount exercise.

Tests taken from [x-common/word-count.json](https://github.com/exercism/x-common/blob/master/word-count.json) and additional tests from JavaScript implementation

### DIFF
--- a/config.json
+++ b/config.json
@@ -12,6 +12,7 @@
     "roman-numerals",
     "robot-name",
     "leap",
+    "word-count",
     "trinary",
     "bowling",
     "clock",
--- a/exercises/word-count/example.php
+++ b/exercises/word-count/example.php
@@ -1,0 +1,18 @@
+<?php
+
+function wordCount($phrase)
+{
+    $count = [];
+
+    $phrase = preg_replace(['/[^a-zA-Z0-9]+/', '/\s\s/'], ' ', mb_strtolower($phrase));
+
+    foreach (explode(' ', $phrase) as $word) {
+        if (!array_key_exists($word, $count) && !empty($word)) {
+            $count[$word] = 1;
+        } elseif (array_key_exists($word, $count)) {
+            $count[$word]++;
+        }
+    }
+
+    return $count;
+}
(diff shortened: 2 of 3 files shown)


petemcfarlane:
Thanks for the PR 👍 I just have a couple of suggestions before merging:
 - If it makes sense for the exercise, we prefer to use functions rather than classes (as per issue #29) so the tests might instead look like `$this->assertEquals(['word' => 1], wordCount('word'));`. This makes sense in this exercise because there is no state to preserve.
 - I would probably not include the `testCountsInternationalCharacters` test, this seems to be conflicting with testing it ignores punctuation, and many of the other language tracks don't include this. I noticed it is in [the javascript version](https://github.com/exercism/xjavascript/blob/9cfb49cd21aba663da308e72580064e95ac941dd/exercises/word-count/word-count.spec.js#L21) but they also test that the returned array __includes__ punctuation, unlike some of the other language tracks.

The rest looks good!

ecrmnn:
I can see how just using a function would make it "easier", but don't you feel a class is more "real world"?
In an "actual project" I wouldn't have a global function called ``wordCount()``, but rather a class.

Also noticed ``Bowling``, ``Clock``, ``Connect``, ``DifferenceOfSquares`` and ``RobotName`` are all classes.

What's your opinion on this?

I agree with you on removing ``testCountsInternationalCharacters``. Just wanted some feedback on it. It adds too much complexity.

petemcfarlane:
I did used to have the same view as you, thinking that using objects would be closer to my day-to-day OO work, but then I became a functional programmer ;-) 

But my serious **_opinion_** is that a function is cleaner to test, less code to write and easier to understand for the person practicing the exercise, and what we're testing in this exercise is how to group/count words, not how to do OOP!

Bowling, Clock and a few others are written in an OO style because they all have __state__. Objects are great for encapsulating state with behaviour or functionality. But in many of these exercises, there is no state, only functionality. In those situations functions are much cleaner and simpler.

ecrmnn:
Refactored to function and removed ``international characters``-test

ecrmnn:
@kytrinyx Ready for merge?

petemcfarlane:
Thanks @ecrmnn, I've now merged it!

PR #79, opened 2016-05-10
Add change probelm and example

petemcfarlane:
Taken from [x-common](https://github.com/exercism/x-common/blob/master/change.md).

I've also added a couple of tests for edge cases:
 - testing for change smaller than the provided coins
 - testing for negative change values
 - testing for different sets of coins
 - testing for large target values

I wonder if I should add this to the x-common `change.md` file?

### DIFF
--- a/config.json
+++ b/config.json
@@ -20,7 +20,8 @@
     "bowling",
     "clock",
     "wordy",
-    "connect"
+    "connect",
+    "change"
   ],
   "deprecated": [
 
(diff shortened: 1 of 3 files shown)


kytrinyx:
These are great edge cases. We could put them in the README, but even more so... how about adding a `change.json`? https://github.com/exercism/x-common/blob/master/change.json - then we can notify all the tracks that implement it that the problem has been updated, and they can pull these edge cases back into the current implementations' test suites.

PR #80, opened 2016-05-25
Remove tests that require modification

brettsantore:
The test suite should not be modified in order for a test to pass. With the exception of skipped tests.

### DIFF
--- a/exercises/gigasecond/gigasecond_test.php
+++ b/exercises/gigasecond/gigasecond_test.php
@@ -64,5 +64,4 @@ class GigasecondTest extends \PHPUnit_Framework_TestCase
 
         $this->assertNotEquals($date, $gs);
     }
-
 }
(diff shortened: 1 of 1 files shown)


petemcfarlane:
I have no strong feelings either way on this, but what do the other language tracks do? In the x-common file it says
```
Finally, as a demonstration but not really a test, some languages demonstrate the add function by inviting the solver to include their birthdate in either the solution code or test program.  The test program then shows or tests their gigasecond anniversay.
```
https://github.com/exercism/x-common/blob/deb225ea46f2c0c20bd240469eda6ef3366f7781/gigasecond.json#L10-L14

Perhaps this should be updated too?

Travis is also whining about the (now) redundant new line (67) at the end of the class.

brettsantore:
I will take a look through x-common

brettsantore:
Some do implement it some don't. Since it is marked as skipped, it should be okay to leave and let the user decide.

kytrinyx:
How about a comment that explicitly says that it's optional? I'm not sure everyone pays particularly good attention to the README.

brettsantore:
The problem I have is that the user has to do the math to have the test pass.

kytrinyx:
Yeah, I removed it from the Go track. If they want to try it, they can, and the README does make the suggestion.

PR #82, opened 2016-05-27
Largest series product: create tests and example

brettsantore:
Adding a new exercise to the track

### DIFF
--- a/exercises/largest-series-product/example.php
+++ b/exercises/largest-series-product/example.php
@@ -1,7 +1,124 @@
 <?php
-/**
- * Created by PhpStorm.
- * User: santore
- * Date: 5/27/16
- * Time: 11:04 AM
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thanks! The problem needs to be added to the config.json file.

Check out the contributing guide section on track-level linting for details:

https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#track-level-linting-with-configlet

brettsantore:
Is there an order to the `config.json?

kytrinyx:
It's very roughly in order of increasing difficulty. If you're not sure where to put it then I would take a look at a couple of the other tracks and see where they put it and then do something similar. e.g. http://exercism.io/languages/go ("available exercises" tab).

PR #83, opened 2016-05-27
Largest series product: create tests and example

brettsantore:
Added the test, an example and updated the `config.josn`

### DIFF
--- a/exercises/largest-series-product/example.php
+++ b/exercises/largest-series-product/example.php
@@ -1,0 +1,117 @@
+<?php
+
+class Series
+{
+
+    /**
+     * The the digits to test
+     *
+     * @var string
+     */
+    private $digitSequence;
+
+    /**
+     * The length of digits
+     *
+     * @var int
+     */
+    private $sequenceLength;
+
+    /**
+     * Series constructor.
+     *
+     * @param string $digits
+     */
+    public function __construct($digits)
+    {
+        $this->validateSequence($digits);
+
+        $this->digitSequence = $digits;
+        $this->sequenceLength = strlen($digits);
+    }
+
+    /**
+     * Return the largest product from the series of digits
+     *
+     * @param int $span
+     * @return int
+     */
+    public function largestProduct($span)
+    {
+        if (0 == $span) {
+            return 1;
+        }
+
+        $this->validateSpan($span);
+
+        return $this->largestSeriesProduct($span);
+
+    }
+
+    /**
+     * Make sure the span is valid
+     *
+     * @param int $span
+     * @throws \InvalidArgumentException
+     */
+    private function validateSpan($span)
+    {
+        if ($span < 0) {
+            throw new InvalidArgumentException(sprintf(
+                '$span must be greater than or equal to 0. [%s] was given',
+                $span
+            ));
+        }
+
+        if ($span > $this->sequenceLength) {
+            throw new InvalidArgumentException(sprintf(
+                '$span can not be greater than the length of the sequence. $span:[%s], sequenceLength:[%s]',
+                $span,
+                $this->sequenceLength
+            ));
+        }
+    }
+
+    /**
+     * Returns the largest product of a string of digits of length $span
+     *
+     * @param int $span
+     * @return int
+     */
+    private function largestSeriesProduct($span)
+    {
+        $products = [];
+        for ($start = 0; $start <= $this->sequenceLength - $span; $start++) {
+            $products[] = $this->multiplyStringSection(substr($this->digitSequence, $start, $span));
+        }
+        return max($products);
+    }
+
+
+    /**
+     * Multiplies a numerical string section
+     *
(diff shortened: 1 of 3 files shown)


kytrinyx:
👍  For what it's worth, you can push new commits to the same branch and it will update the old pull request. If you squash or rebase and the branch refuses to accept the push, then you can `git push -f` (for --force).

brettsantore:
Thank you. It's easy to work on your own projects, but I need some assistance helping with others.

brettsantore:
should i have kept the same commit message?

kytrinyx:
Nah, as long as the final commit message makes sense when people are perusing the history, it's fine.

I probably would have gone with something like `Implement largest-series-product` or `Add largest-serie-product exercise`... but really as long as we have the problem name and can figure out what happened, it's all good.

I'm leaving this for @petemcfarlane to review.

brettsantore:
I will keep this in mind for my future additions

exercises/largest-series-product/example.php
```diff
@@ -0,0 +1,124 @@
...
+        $this->validateSequence($digits);
+
+        $this->digitSequence = $digits;
+        $this->sequenceLength = strlen($digits);
+    }
+
+    /**
+     * @return int
+     */
+    public function getSequenceLength()
```
petemcfarlane: I don't think you need this method at all, or am I missing something? It shouldn't be public at least. Where you have called this couldn't you just use the property, e.g. line 60 would be come `if ($span > $this->sequenceLength) {`

exercises/largest-series-product/example.php
```diff
@@ -0,0 +1,124 @@
...
+     */
+    private function multiplyStringSection($stringSection)
+    {
+        return array_product(str_split($stringSection));
+    }
+
+    /**
+     * @param string $digits
+     */
+    public function setSequence($digits)
```
petemcfarlane: I would remove this function and just move the three calls into the constructor. To me it seems like a smell if the only call in the constructor is delegating to this method. It seems unnecessary to do so in this case.

exercises/largest-series-product/example.php
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    /**
+     * @param $span
+     * @return int|number
+     */
+    private function largestSeriesProduct($span)
+    {
+        $products = [];
```
petemcfarlane: More of a stylistic improvement than changing behaviour, but has you thought about using an `array_map` here, for example
```php
$products = array_map(function ($start) use ($span) {
    return $this->multiplyStringSection(substr($this->digitSequence, $start, $span));
}, range(0, $this->sequenceLength - $span));
```

petemcfarlane:
Thanks for the contribution @brettsantore 😄

exercises/largest-series-product/example.php
brettsantore: In this case, it does make sense to move it to the constructor. I was attempting to give it some flexibility.

exercises/largest-series-product/example.php
brettsantore: I don't feel strongly toward either approach. Since this is the heart of the problem it should be open to some different approaches.

exercises/largest-series-product/example.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        $this->validateSequence($digits);
+
+        $this->digitSequence = $digits;
+        $this->sequenceLength = strlen($digits);
+    }
+
+    /**
+     * Return the largest product from the series of digits
+     * @param $span
```
petemcfarlane: Could you inject a newline before the `@param` for consistency? And what does number in `@return int|number` mean? Would `@return int` be enough?

exercises/largest-series-product/example.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    private function multiplyStringSection($stringSection)
+    {
+        return array_product(str_split($stringSection));
+    }
+
+    /**
+     * @param $digits
+     */
+    private function validateSequence($digits)
```
petemcfarlane: It makes sense to annotate this exception as you have on `validateSpan` with `@throws \InvalidArgumentException`

exercises/largest-series-product/example.php
brettsantore: `int|number` was an artifact from phpstorm's auto doc generator

brettsantore:
I jumped the gun with the last commit, i'm going through to make all the doc's consistent

petemcfarlane:
Thanks!

brettsantore:
Thanks for bearing with me, I'll make sure to go through similar things with my future pull requests.

petemcfarlane:
not at all, I'm being picky as this is an _exemplar_ code sample, so should ooze best practices and quality to any one looking at it for reference! Thanks for the contribution, feel free to do any more you may fancy.

PR #85, opened 2016-05-31
Add bob excersize

brettsantore:
Add test, and example and line in place to config file.

### DIFF
--- a/exercises/bob/bob_test.php
+++ b/exercises/bob/bob_test.php
@@ -1,0 +1,205 @@
+<?php
+require_once "bob.php";
+
+class BobTest extends PHPUnit_Framework_TestCase
+{
+
+    /**
+     * @var \Bob
+     */
+    private $bob;
+
+    public function setUp()
+    {
+        $this->bob = new Bob;
+    }
+
+    public function testStatingSomething()
+    {
+        $this->assertEquals("Whatever.", $this->bob->respondTo("Tom-ay-to, tom-aaaah-to."));
+    }
+
+    public function testShouting()
+    {
+        $this->markTestSkipped();
+
+        $this->assertEquals("Whoa, chill out!", $this->bob->respondTo("WATCH OUT!"));
+    }
+
+    public function testShoutingGibberish()
+    {
+        $this->markTestSkipped();
+
+        $this->assertEquals("Whoa, chill out!", $this->bob->respondTo("FCECDFCAAB"));
+    }
+
+    public function testAskingAQuestion()
+    {
+        $this->markTestSkipped();
+
+        $this->assertEquals("Sure.", $this->bob->respondTo("Does this cryogenic chamber make me look fat?"));
+    }
+
+    public function testAskingANumericQuestion()
+    {
+        $this->markTestSkipped();
+
+        $this->assertEquals("Sure.", $this->bob->respondTo("You are, what, like 15?"));
+    }
+
+    public function testAskingGibberish()
+    {
+        $this->markTestSkipped();
+
+        $this->assertEquals("Sure.", $this->bob->respondTo("fffbbcbeab?"));
+    }
+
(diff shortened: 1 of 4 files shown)


brettsantore:
I don't love the class knowing about the unicode characters, but I'm not sure about a solution.

It's either breaking PSR-1 or starting to use multiple includes, since we can't really rely on composer, or efficiently implement PSR-4 autoloading.

 Thoughts?

exercises/bob/bob_test.php
```diff
@@ -0,0 +1,203 @@
...
+{
+
+    /**
+     * @var \Bob
+     */
+    private $bob;
+
+    public function setUp()
+    {
+        $this->bob = new Bob(new Unicode());
```
petemcfarlane: Is `new Unicode()` used for anything?

exercises/bob/bob_test.php
```diff
@@ -0,0 +1,203 @@
...
+    }
+    public function testStatingSomething()
+    {
+        $this->assertEquals("Whatever.", $this->bob->respondTo("Tom-ay-to, tom-aaaah-to."));
+    }
+
+    public function testShouting()
+    {
+//        $this->markTestSkipped();
```
petemcfarlane: I think you forgot to uncomment the `markTestSkipped()` in quite a few places ;)

exercises/bob/example.php
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+
+    /**
+     * Test if the string is being yelled
+     *
+     * @param string $str
+     * @return bool
+     */
+    private function isYelling($str)
```
petemcfarlane: `isYelling` sounds a lot like `isShouting` to me. Perhaps a better name might be `isUppercase`?

petemcfarlane:
I'm wondering if adding the `testShoutingWithUmlauts` (`ÜMLÄÜTS!`) and `testOtherWhitespace` (`\n\r \t\u000b\u00a0\u2002`) should be reserved as stretch goals. They are currently in [x-common](https://github.com/exercism/x-common/blob/1543b4536911bd5bc2995e81868b59897eb98478/bob.json). The first requires the `mb_string` extension which may not be installed by default (as mentioned in issue https://github.com/exercism/xphp/issues/81), and the second is just a bit hacky it seems. Do you have an opinion @kytrinyx?

exercises/bob/bob_test.php
brettsantore: This was an artifact from another approach. I removed it, but I'm also confused why no tests picked that up as a problem. Everything passes…

exercises/bob/example.php
brettsantore: I have addressed this.

brettsantore:
I would agree, but I'll wait to push any changes.

exercises/bob/bob_test.php
petemcfarlane: I thought it might have been. You have no `__construct()` method defined so whatever you pass will be ignored, just like if you provide too many parameters to a method or function they are ignored.

kytrinyx:
Agreed: Wherever possible I think it's good to reduce the number of external dependencies. Let's make those stretch-goals.

brettsantore:
So I made the changes and commented out the stretch goals.

I would propose extending test cases for stretch goals in some way. `class StretchBobTest extends BobTest`,

petemcfarlane:
That looks good, could we just add a `HINTS.md` file which very briefly explains
> there are some extra tests at the end of the file which are commented out because they are stretch goals. They may be easier to solve if you are using the `mb_string` functions, which isn't installed by default with every version of PHP.

or something like that?

brettsantore:
let me know what you think

petemcfarlane:
I like it, thanks 😄

PR #86, opened 2016-05-31
Add pig-latin exercise

brettsantore:
Added a test, and example, and update the `config.json`

### DIFF
--- a/config.json
+++ b/config.json
@@ -14,6 +14,7 @@
     "difference-of-squares",
     "largest-series-product",
     "roman-numerals",
+    "pig-latin",
     "robot-name",
     "leap",
     "word-count",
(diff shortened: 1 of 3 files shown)


petemcfarlane:
Looks good but you've left the `markTestSkipped()` commented out in the test file.
Thanks

brettsantore:
Fudge, 2nd time I've done that.

petemcfarlane:
No worries. If it's easier than uncommenting and renaming the files, I tend to run PHPUnit using the `make` command from the xphp directory, like so 
```
make test-assignment ASSIGNMENT=pig-latin
```
This takes care of the renaming and removing the `markTestSkipped` comments too, as seen in the Makefile.

PR #92, opened 2016-06-20
Add minesweeper exercise

petemcfarlane:


kytrinyx:
Hi Pete, at first blush this looks good to me. Would you mind squashing the commits?

I noticed that this exercise doesn't have a canonical data set at the moment. Did you base this on a different implementation? If so, which one?

petemcfarlane:
Yup I was going to use the GH Merge & Squash feature when ready, but I've done it the good old fashioned way now.

I did it as a kata with a colleague. We took inspiration from the [Ruby](https://github.com/exercism/xruby/blob/0402d11ab6567c16597639597e9aa92351c60f55/exercises/minesweeper/minesweeper_test.rb) implementation, but when we tried to TDD this we found some of the tests to be incomplete or out of order. Their test suite didn't lend itself to incrementally passing the tests, but just expected everything to work.

They also changed their inputs/outputs to a list of strings, rather than just multi-line strings. I don't suppose this is much of an issue but the [original](https://github.com/exercism/x-common/blob/b89aec5cd43cd246350b5d52cccb813002d2a0bc/minesweeper.md) just compares strings.

Is there an easy way to filter the languages which have implemented an exercise, so people can compare when writing a new exercise? Do you have any comments on this test suite, or would you like me to submit a `minesweeper.yml` to the x-common branch?

kytrinyx:
> I was going to use the GH Merge & Squash feature when ready, but I've done it the good old fashioned way now.

I forgot about that feature. That would have been perfectly good :)

> we tried to TDD this we found some of the tests to be incomplete or out of order.

Interesting! I wonder if that's the case in Ruby as well, in which case this could be a good improvement to suggest there.

> They also changed their inputs/outputs to a list of strings, rather than just multi-line strings.

Yeah, I don't have strong opinions about that.

> Is there an easy way to filter the languages which have implemented an exercise, so people can compare when writing a new exercise?

Yepp. We have this endpoint: http://x.exercism.io/v3/problems/minesweeper/todo (look for "implementations").

> Do you have any comments on this test suite

No, I think this is good to merge.

> would you like me to submit a minesweeper.yml to the x-common branch

You mean minesweeper.json? I think that's a completely separate problem. I wouldn't worry about that here.

petemcfarlane:
>Yepp. We have this endpoint: http://x.exercism.io/v3/problems/minesweeper/todo (look for "implementations").

Ah I didn't know about that endpoint, but that's really useful to know, thanks!

Glancing at a few other languages implementations, many of them don't have a border around their game, and don't have as much error checking. Removing these tests would significantly clean up the test and example and make the challenge quite a bit more straightforward.

Do you think it is fine as it is? The tests do cover the edge cases, and it is how we ended up test-driving the implementation, but we'd reduce some edge cases and coding if we removed borders and/or malformed inputs...

kytrinyx:
I think it's fine as it is. If/when someone starts tackling the minesweeper.json canonical test data, we can have a more general discussion about what makes this a good exercise, and how detailed the requirements should be, etc.

PR #93, opened 2016-07-02
Fix clock_test.php method calls

nhoag:
Resolves the following error when executing `phpunit clock_test.php`:

```
PHP Fatal error:  Call to a member function __toString() on null in /path/to/exercism/php/clock/clock_test.php on line 27
```


nhoag:
Aha! It can be solved as-is. Closing.

petemcfarlane:
no worries 👍

PR #99, opened 2016-07-27
New about

yesroh:
Making some changes, forgiveness for the English. It is not my mother tongue. I am open to advice and suggestions #96 #97

PR #100, opened 2016-08-02
Add ABOUT.md

yesroh:
suggested corrections on the  PR #99


docs/ABOUT.md
```diff
@@ -1,6 +1,6 @@
...
+PHP is a development language intended for the web that has become popular because it is easy to learn, over time it has become a modern language and so won its space in the roll of the most respected programming languages. It was originally created by Rasmus Lerdorf in 1994 to make it easier to manage a personal website, today it is used to many other things like the social networking [Facebook](https://www.facebook.com/), the open-source tool to create blogs and sites (WordPress)[https://wordpress.com/create/], the e-commerce open-source plataform (Magento)[https://magento.com/], and many other thins on web. Getting started with PHP is very simple! Just add some PHP-tags to an already existing HTML file and voila!
```
petemcfarlane: I see you fixed the links for Facebook, looks like you missed the Wordpress and Magento ones. The format is`[the link text](the url)`. (I always get these mixed up myself too!)

petemcfarlane: A couple of minor typos: 

- `thins on web` should be `things on the web`.
- `today it is used to many other things ` could be `today it is used for many other things `
- `the open-source tool to create blogs and sites (WordPress)` how about `the open-source blogging tool [WordPress](...)`

Thanks for doing this! When you're done editing you can just commit the file, and `git push` back up to the _same_ branch. That way GitHub will update the Pull Request magically and you won't have to create another! It also helps us to keep all the feedback and conversation within the same Pull Request.
I'll close PR #100 and #99 now.

docs/ABOUT.md
```diff
@@ -0,0 +1,7 @@
+# PHP
```
kytrinyx: We don't need the title in the ABOUT.md, since it gets included into a page that already has a title.

kytrinyx:
Since we've not seen any commits from the original author I'm going to go ahead and close this.

Thank you for your contribution, @yesroh! If you decide to come back, make sure to check what's in master and then work off the most recent version of the file.

PR #101, opened 2016-08-07
Add Phone Number problem

nhoag:


nhoag:
Passed!

exercises/phone-number/phone-number_test.php
```diff
@@ -0,0 +1,147 @@
...
+{
+    public function testCleansNumber()
+    {
+        $number = new phoneNumber('(123) 456-7890');
+        $this->assertEquals('1234567890', $number->number());
+    }
+
+    /**
+     * @depends testCleansNumber
+     */
```
petemcfarlane: Hey, do you think we need these `@depends ...` annotations if we're using `markTestSkipped()` already? I think they are redundant, what do you think?

exercises/phone-number/phone-number_test.php
```diff
@@ -0,0 +1,147 @@
+<?php
+
+require "phone-number.php";
+
+class PhoneNumberTest extends \PHPUnit_Framework_TestCase
+{
+    public function testCleansNumber()
+    {
+        $number = new phoneNumber('(123) 456-7890');
```
petemcfarlane: Would you mind making `phoneNumber` `PhoneNumber` instead, as it's common for PHP class names to use [Studly case](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#appendix-a-survey)

exercises/phone-number/phone-number_test.php
```diff
@@ -0,0 +1,147 @@
...
+        $this->assertEquals('1234567890', $number->number());
+    }
+
+    /**
+     * @depends testCleansNumber
+     */
+    public function testCleansADifferentNumber()
```
petemcfarlane: I would say all of the tests with the word _different_ in them are redundant and just testing the same method with different input data. Whilst they may be helpful for doing TDD I'd probably remove these redundant test cases to clean up the file a bit, as we're not really adding any new behaviour in any of the cases, only making the test file longer to read. Would you agree?

petemcfarlane:
Hey @nhoag thanks for the PR! Let me know if you agree/disagree with any of my comments and we'll move to getting it merged. Cheers 👍

exercises/phone-number/phone-number_test.php
nhoag: I kind of liked the behavior of `@depends` from the 'connect' exercise, but they are definitely redundant with `markTestSkipped()`. Plus `markTestSkipped()` is consistent with a majority of exercises. Removed `@depends`!

exercises/phone-number/phone-number_test.php
nhoag: Good catch! I changed this in `example.php`, but not here. Updated.

exercises/phone-number/phone-number_test.php
nhoag: Agreed. Removed!

nhoag:
@petemcfarlane Thanks for the feedback!

petemcfarlane:
Lovely, thanks!

PR #102, opened 2016-08-09
Add missing testcase for 9 (3 x 3) in raindrops ex

kenden:
In Go, Javascript, Ruby, Rust tracks, this testcase is present.


petemcfarlane:
Good spot, thanks!

PR #103, opened 2016-08-13
Added verbose names to tests in gigasecond_test.php

tstirrat15:
Also added a comment that explains the `your_birthday` test. It doesn't run out of the box.


petemcfarlane:
I really like this improvement, thanks 👍

PR #104, opened 2016-08-16
Update LEARNING.md

allyjweir:
Add online learning resources links.

I made sure to only include free learning resources to lower any barrier of entry an individual might experience. Some of these resources may not always be free so may need updated in the future as required.


kytrinyx:
Thank you!

kytrinyx:
> I made sure to only include free learning resources to lower any barrier of entry an individual might experience

I think that's a good call.

PR #107, opened 2016-08-23
Add initial 'exercises' key to config.json (for issue #105)

allyjweir:
Work based on issue #105. 

Have only added difficulties to the exercises that I have completed so far. I would be happy for it to be merged as is but it would be better if someone else could contribute difficulty ratings for the rest of the exercises.

If there is a way that we can get others to add commits to this PR that would be ideal but unfamiliar with process. If anyone is inclined to add to it, please let me know of anything I need to do to facilitate that.

### DIFF
--- a/config.json
+++ b/config.json
@@ -159,6 +159,6 @@
     "slug": "phone-number",
     "difficulty": 1,
     "topics": []
-  },
+  }
 ]
 }
--- a/config.json
+++ b/config.json
@@ -38,5 +38,127 @@
   ],
   "foregone": [
 
-  ]
+  ],
+  "exercises": [
+  {
+    "slug": "hello-world",
+    "difficulty": 1,
+    "topics": []
+  },
+  {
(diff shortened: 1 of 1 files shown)


petemcfarlane:
Thanks @allyjweir. I don't know of an easy way to let multiple people contribute to a branch especially when it is forked to your own GitHub namespace. I think I will merge the work you have done so far into master, then others can contribute to set the topics and difficulty ratings. 

I feel like the difficulty ratings will be quite subjective, but I think your representation is quite fair. These should be open for debate and discussion if other people have different opinions and ideas, but they are a great starting point!

kytrinyx:
I believe you can add others as collaborators to your fork under [Settings > Collaboration](https://github.com/allyjweir/xphp/settings/collaboration).

allyjweir:
Will remember that for next time @kytrinyx, thanks.

PR #109, opened 2016-09-20
Add Atbash Cipher exercise

nhoag:


petemcfarlane:
Hey thanks for submitting another exercise! I think this could be better/simpler to just be a function rather than a class with a single method. It would also clear the test up some. 

`$this->assertEquals('ml', atbashEncode('no'));` or simply `encode('no')`

I think it is simpler because there is no concept of state, it is a pure function. For any input the output will always be the same with no side effects. What do you think @nhoag?

nhoag:
Updated! That sounds very sensible. Thanks for the review!

petemcfarlane:
Great, thanks!

PR #110, opened 2016-09-27
Add binary exercise.

pmatseykanets:


exercises/binary/binary_test.php
```diff
@@ -0,0 +1,102 @@
...
+        $this->setExpectedException(InvalidArgumentException::class);
+        parse_binary('2');
+    }
+
+    public function test_id_dosnt_accept_5()
+    {
+        $this->markTestSkipped();
+
+        $this->setExpectedException(InvalidArgumentException::class);
+        parse_binary('2');
```
petemcfarlane: I think this should be a 5?

exercises/binary/binary_test.php
```diff
@@ -4,94 +4,94 @@
...
-    public function test_binary_10001101000_is_decimal_1128()
+    public function testBinary10001101000IsDecimal1128()
     {
         $this->markTestSkipped();
 
         $this->assertEquals(1128, parse_binary('10001101000'));
     }
 
-    public function test_id_dosnt_accept_2()
+    public function testItDosentAccept2()
```
petemcfarlane: There are quite a lot of negative test examples - `testItDoesntAccept...` These might be useful to TDD your code but I would probably generalise these into one test case, e.g. `testItOnlyAcceptsCombinationsOfZeroAndOne()`

pmatseykanets:
I prefer snake case in tests too. The reason I changed it to camelCase is because the code sniffer was killing the travis build with 

```
FILE: /home/travis/build/exercism/xphp/exercises/binary/binary_test.php
--------------------------------------------------------------------------------
FOUND 13 ERRORS AFFECTING 13 LINES
--------------------------------------------------------------------------------
  7 | ERROR | Method name "BinaryTest::test_binary_1_is_decimal_1" is not in
    |       | camel caps format (PSR1.Methods.CamelCapsMethodName.NotCamelCaps)
 12 | ERROR | Method name "BinaryTest::test_binary_10_is_decimal_2" is not in
    |       | camel caps format (PSR1.Methods.CamelCapsMethodName.NotCamelCaps)
...
```
 I'd love to revert it back if you reconfigure code sniffer at least for test files.

exercises/binary/binary_test.php
pmatseykanets: Absolutely. Good catch. Thanks!

exercises/binary/binary_test.php
pmatseykanets: Good point.

petemcfarlane:
Ah, of course, the code sniffer won't allow method names with underscores! In this case, I don't think it really matters, I think consistency is king and it may be harder or confusing for newcomers to have a standards for the test cases, and a different standard for the implementation example. But if you feel strongly for underscores in test names then make an issue/pr and we can review it by all means!
Thanks

pmatseykanets:
Refactored tests.

Please let me know if anything else needs to be done before merging it in.

Thanks.

PR #111, opened 2016-09-27
Add accumulate exercise

pmatseykanets:


exercises/accumulate/accumulate_test.php
```diff
@@ -0,0 +1,111 @@
...
+        $this->markTestSkipped();
+
+        $accumulator = function ($string) {
+            return strrev($string);
+        };
+
+        $this->assertEquals(['Hello', 'World!'], accumulate(['olleH', '!dlroW'], $accumulator));
+    }
+
+    public function testAccumulateContants()
```
petemcfarlane: should that be `testAccumulateConstants()`?

exercises/accumulate/accumulate_test.php
```diff
@@ -0,0 +1,112 @@
+<?php
+
+require_once 'accumulate.php';
+
+// These required for last three tests
+require_once 'test_stub_static.php';
```
petemcfarlane: Requiring other files (other than the example) isn't going to pass the build as we use the [Makefile](https://github.com/exercism/xphp/blob/a599c251a6a897ff00b2b427ef5df0e438a6e16d/Makefile#L25-L29) to copy the example and the test to a new test dir. You could just add these classes inline to the bottom of the test file - that would then pass

pmatseykanets: That is exactly how it was done it the first place (mock classes at the bottom of the test). But again, like in #110, the code sniffer was complaining that every class should go in its own file (which is how it should be but not for tests) and killing travis builds.

petemcfarlane: Oh for goodness sake! Hmm, maybe we need to have a look at the code sniffer rules. Lets raise it as an issue

pmatseykanets:
I reverted the commits that came with the binary exercise and committed the proposed change in #113 in order to make sure that we're all green. 

Please let me know if anything else should be done before it's merged.

Thanks.

petemcfarlane:
Hi @pmatseykanets this now looks great but I can't merge it because it will conflict with the other recent merges - `config.json` is conflicting. Please could you rebase/merge the master branch, then push that up? Thanks!

pmatseykanets:
@petemcfarlane I reconciled the upstream. Please merge.

Thanks.

petemcfarlane:
Thanks!

PR #117, opened 2016-10-04
Add Variable length quantity exercise

pmatseykanets:


exercises/variable-length-quantity/variable-length-quantity.php
```diff
@@ -0,0 +1,18 @@
...
+ * @return array
+ */
+function vlq_encode($int)
+{
+    // YOUR CODE GOES HERE
+}
+
+/**
+ * @param array $bytes
+ */
```
petemcfarlane: It's not common for the exercises to have a skeleton file to get the user started (other than the Hello World exercise), so you could remove this, but I have no strong feelings. However, if you keep it then can we please add a `@return` type flag to this for consistency?

petemcfarlane:
Great thanks!

pmatseykanets:
Sure! One thing that I didn't ask is what difficulty level to put for this exercise so I left it as 1. Let me know if this needs to change.

Thanks!

petemcfarlane:
Ok, to be honest I don't think any of the difficulties are set yet, I think that could be another issue. If you were to rate it out of 10, compared with other exercises, where do you think it would be? I'd have probably said somewhere between 5 and 10

pmatseykanets:
I'd say 5 is fair.

pmatseykanets:
@petemcfarlane I'm working on the canonical-data.json for this exercise and I have a few questions. Is there a way to express a test with multiple assertions using canonical-data schema? For example like this one

```php
    public function testItEncodesSingleBytes()
    {
        $this->assertEquals([0x00], vlq_encode([0x00]));
        $this->assertEquals([0x40], vlq_encode([0x40]));
        $this->assertEquals([0x7f], vlq_encode([0x7f]));
    }
```

Also I see across different `canonical-data` files that properties expressing input and expected values are called differently. Is there any preferred/proper way of naming them?

Thanks.

petemcfarlane:
This I don't know, I've never done it myself, only read them! Maybe @kytrinyx can point you in the right direction though?

pmatseykanets:
Hi @kytrinyx are there any pointers? Thanks.

PR #121, opened 2016-10-17
instantiate classes, add test case

woylie:
The tests will now properly instantiate classes instead of calling PigLatin::translate(); directly. This allows the usage of member variables etc. in solutions. I've also added a test case as proposed in another pull request to x-common to check for proper implementation of the consonant+qu rule.

### DIFF
--- a/exercises/pig-latin/example.php
+++ b/exercises/pig-latin/example.php
@@ -30,4 +30,3 @@ function translateWord($str)
 
     return sprintf("%s%say", substr($str, 1), $str[0]);
 }
-
--- a/exercises/pig-latin/example.php
+++ b/exercises/pig-latin/example.php
@@ -1,36 +1,33 @@
 <?php
-
-class PigLatin
+/**
+ * Translates a string into Pig Latin
+ *
+ * @param  string $str
+ * @return string
+ */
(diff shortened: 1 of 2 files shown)


woylie:
Hi @petemcfarlane, I thought it would be better because you might want to break your function up into multiple functions. If you do so, you would also have to call PigLatin::myFunction() from within the class instead of calling $this->myFunction(). That's why I changed it. If you choose to use a class I think you should use it as you normally would.

However, a plain function would be fine as well in this case. Do you want me to change it?

petemcfarlane:
Hi @woylie. I think I see what you mean about breaking methods up into multiple, smaller functions, but I disagree that calling the private static methods would be something you wouldn't _normally_ do. Using private static methods like this is perfectly valid PHP, especially on an object that doesn't have state, only has pure functions or is immutable.

I think to make the exercise as approachable as possible, and not have to worry about classes, static or not-static methods, private/public methods, we should probably just make it a function.

Are you sure you wouldn't mind updating the PR? I'm happy to help if you don't have the time.

Thanks!

woylie:
Hi @petemcfarlane, I just made the change.

PR #136, opened 2016-11-29
Add tests and example for Sieve exercise

jeslopcru:
Related with PR https://github.com/exercism/xphp/pull/135
Add tests and example for Sieve


exercises/sieve/sieve_test.php
```diff
@@ -0,0 +1,201 @@
+<?php
+require "sieve.php";
+class SieveTest extends PHPUnit_Framework_TestCase
+{
+    public function testNoPrimesUnderTwo()
+    {
+        $this->assertEquals([], sieve('1'));
```
petemcfarlane: This looks great, thank you for resubmitting the PR. The only think I'd like to change is `sieve('1')` to `sieve(1)`, and for the other tests too. I know PHP will coalesce strings to numbers when it can (providing the user isn't using strict mode). Since we're doing arithmetic I think an int would be better than a string :)
Thanks!

petemcfarlane:
Perfect, thanks!

PR #137, opened 2016-12-02
Names uniqueness tests (ported from GO version)

Hrumpa:
Tests in PHP version of this task lacks the MOST interesting part - checking for uniqueness of new names. Ported tests from GO version.

### DIFF
--- a/exercises/robot-name/example.php
+++ b/exercises/robot-name/example.php
@@ -1,37 +1,76 @@
 <?php
-
 class Robot
 {
-    protected $name = null;
-
-    protected $alphabet = '';
+    private $name;
 
     public function __construct()
     {
-        $this->alphabet = array_flip(str_split('ABCDEFGHIJKLMNOPQRSTUVWXYZ'));
+        $this->reset();
     }
 
+    /**
+     * Get Robot name
+     *
+     * @return string
+     */
     public function getName()
     {
-        if (is_null($this->name)) {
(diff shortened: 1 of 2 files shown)


petemcfarlane:
Hey @Hrumpa, do you have any suggestion how we might enforce this in the example? It seems we'd need to store a list of the previously used robot names somewhere. Would you like to submit an update to the `example.php` file? Thanks!

Hrumpa:
Now it's OK, was a bit messy with PSR. As I've said before, the real purpose of the task should be outlined a bit better, if it is not about inventing new GUID with at least 10000 honest random ids in the small max of 676000 (will gladly take a look at the solution).

kytrinyx:
A couple of small things before I leave this to @petemcfarlane for final review:

Would you mind rebasing and squashing all the commits into a single commit? All of these things are about the same change, so it doesn't make sense to have many commits in this case.

Second, would you make sure that the summary line of the commit message contains `robot-name` (the name of the exercise)? That makes it much easier when we're searching the logs for things.

If you have more details (such as the fact that you ported the tests from Go) these could go in the body of the commit message.

My favorite article about git commit messages is this one: http://chris.beams.io/posts/git-commit/
 in case you like reading about this stuff.

petemcfarlane:
Thanks @Hrumpa, I'll get this merged if you could rebase/squash the commits as @kytrinyx mentioned!

Hrumpa:
Is it ok now?

petemcfarlane:
Great, thanks @Hrumpa! 👍

PR #140, opened 2016-12-22
Variable-length-quantity: enhance testOverflow

Hrumpa:
1. 64-bit systems PHP_INT_MAX is 9223372036854775807. Check may be done against PHP_INT_MAX. Or the MAX possible outcome value should be outlined in the task description.
2. Encoded sequence should be completed with valid last item. Depending on realization validation of the last item may precede anything else and throw InvalidArgumentException(). Overflow and invalid input should not be mixed in one test.


petemcfarlane:
Thanks @Hrumpa!

PR #141, opened 2016-12-30
Add Grains exercise

Hrumpa:
I believe that the task may be set up a bit more interesting for the PHP implementation.
 
PHP integers greater than 2^31 (32-bit systems) or 2^63 (64-bit) are casted to floats. In some cases it may lead to problems:
* http://php.net/manual/ru/language.types.float.php#warn.float-precision

Consider King hates floats and demands solution with precise integer-only calculations. Don't involve any floats. Don't use gmp or any other similar libraries. Try to make the solution for virtually any board size.

### DIFF
--- a/config.json
+++ b/config.json
@@ -307,6 +307,14 @@
     "slug": "markdown",
     "difficulty": 3,
     "topics": ["refactoring"]
+  },
+  {
+    "slug": "grains",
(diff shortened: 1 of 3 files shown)


petemcfarlane:
Thanks for submitting this exercise!

PR #143, opened 2017-01-17
Add OCR Numbers exercise

Hrumpa:


exercises/ocr-numbers/example.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Translate OCR to digit
+     * @return string
+     */
+    public function getDigit()
+    {
+        if ($this->digit === null) {
+            $encoded = str_replace(' ', 'x', implode($this->ocr));
```
petemcfarlane: Is there any reason for encoding spaces as `x`?

Hrumpa: The only reason is that I always prefer to keep array keys names along with terms of PHP variable naming rules. If you are not happy with this, let's stay with the bare `implode` and change the mapping.

petemcfarlane: Ah it was more a curiosity than requesting you change it. The examples are mostly there just to get the tests to pass, and for those curious enough to peek at this GitHub repo, so as long as the tests are passing and the test case is sound I'm happy to merge 👍

PR #158, opened 2017-05-30
Add watcher script to exercises.

dkinzer:
I'd like to share this script that I use to run the test automatically as
I make changes to either the test file or the source file.  It's quite helpful
at giving immediate feedback.  Just run `bash watcher.sh`

### DIFF
--- a/exercises/accumulate/watcher.sh
+++ b/exercises/accumulate/watcher.sh
@@ -1,0 +1,23 @@
+# This watcher script works for exercism PHP tests.
+# Run using `bash watcher.sh`
+
+if [ -z "$1" ]; then
+  FILE=$(ls *_test.php | sed -e "s/_test.php//g")
+else
+  FILE=$1
+fi
+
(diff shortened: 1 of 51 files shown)


kytrinyx:
@dkinzer Thanks so much for sharing!

In general my philosophy is to stick with vanilla tools as much as possible, and let people discover their own tweaks and improvements as they go. From that perspective I'd rather not add this directly to the repo and the exercises.

That said, I think that this is something that people might enjoy learning about. How about writing a blog post that we could link to from the resources?

@petemcfarlane What do you think?

dkinzer:
@kytrinyx, yeah that makes total sense to me.  cheers.

petemcfarlane:
Agreed, I think this approach adds too much boiler plate and non-related bloat, but file watchers are very useful. The ones I use all the time are the node `watch` package (not just for javascript files), `fswatch`, and the one baked into PHPStorm IDE.

PR #173, opened 2017-10-05
Exercise Collatz Conjecture

kk-r:


exercises/collatz-conjecture/collatz-conjecture_test.php
```diff
@@ -0,0 +1,44 @@
...
+
+class CollatzConjecture extends PHPUnit\Framework\TestCase
+{
+
+    public function testZeroStepsForOne()
+    {
+        $this->assertEquals(0, steps(1));
+    }
+
+    public function testFourStepsForSixteen()
```
petemcfarlane: This test name is very specific, and it duplicates what is written in the body of the assertion. Can we use the original generic, descriptions, as found in the data file [here](https://github.com/exercism/problem-specifications/blob/b7a941374404249756173d4a4b1575fad77da134/exercises/collatz-conjecture/canonical-data.json#L12), something like `testDivideIfEven()`. It's a more generic definition and will help lead the user to solve the deeper problem.

petemcfarlane: The same for all the other test names, please!

exercises/collatz-conjecture/example.php
```diff
@@ -0,0 +1,17 @@
...
+{
+    $stepCount = 0;
+    if ($number < 1) {
+        throw new InvalidArgumentException('Only positive numbers are allowed');
+    }
+    do {
+        if ($number === 1) break;
+        $stepCount++;
+        $number = ( $number % 2 === 0 ? $number / 2 : 3 * $number + 1);
+    } while($number !== 0);
```
petemcfarlane: There's a couple of linting issues with this example, details from Travis can be [seen here](https://travis-ci.org/exercism/php/builds/285428061#L528-L538) namely, 

- [ ] missing space after `while`
- [ ] need curly braces after the `if` clause
- [ ] one to many blank lines at the end of the file

kk-r:
@petemcfarlane I changed code and pushed back. Please check. :)

PR #180, opened 2017-10-19
Implementing the All Your Base exercise

camilopayan:
Calling dibs on this one!

### DIFF
--- a/exercises/all-your-base/all-your-base_test.php
+++ b/exercises/all-your-base/all-your-base_test.php
@@ -1,0 +1,135 @@
+<?php
+
+require "all-your-base.php";
+
+/**
+ * These tests are adapted from the canonical data in the
+ * `problem-specifications` repository.
+ */
+class AllYourBaseTest extends PHPUnit\Framework\TestCase
+{
+    public function testSingleBitOneToDecimal()
+    {
+        $this->assertEquals(rebase(2, [1], 10), [1]);
+    }
+
+    public function testBinaryToSingleDecimal()
+    {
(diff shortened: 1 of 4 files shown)


camilopayan:
Ready for review!

exercises/all-your-base/all-your-base_test.php
```diff
@@ -0,0 +1,177 @@
...
+ */
+class AllYourBaseTest extends PHPUnit\Framework\TestCase
+{
+    public function testSingleBitOneToDecimal()
+    {
+        $this->assertEquals(rebase(2, [1], 10), [1]);
+    }
+
+    /**
+     * @markTestSkipped
```
petemcfarlane: The tests look fine, I'm wondering if we should change these from annotations to be `$this->markTestSkipped()` for consistency with the other exercises - what are your thoughts @camilopayan ?

exercises/all-your-base/all-your-base.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+function rebase(int $fromBase, array $digits, int $toBase)
```
petemcfarlane: It's not normally necessary to include the answer file with the exercise. The only exception to that is `bob`, which is the first exercise to get people start. So this file can be removed please 👍

exercises/all-your-base/all-your-base_test.php
camilopayan: I've always used the annotation, but changing it shouldn't be a big deal.

PR #186, opened 2017-10-31
Added Codacy badge

borgogelli:
Why don't add the Codacy badge ? Static analysis could be instructive.


petemcfarlane:
Sure, I've not used Codacy but looks pretty cool

PR #188, opened 2017-11-13
Implement Run Length Encoding

Leprechaunz:


Leprechaunz:
@petemcfarlane Makes sense to me, so I removed class to simplify and fixed checkstyle.
But according to Travis I used wrong UUID in config.js (actually I copied it from Python version).
Could you please give me a hint how can I generate correct UUID?

Leprechaunz:
@petemcfarlane Looks like I figured out how to set unique UUID

PR #191, opened 2017-12-27
Update multi-byte character test

### LINKED ISSUE #190: Bug in multi-byte test case for PHP/Bob
Originally reported by @MattLewin 

-----

Currently, the test case reads:
```
    public function testOtherWhitespace()
    {
        $this->markTestSkipped();

        $this->assertEquals("Fine. Be that way!", $this->bob->respondTo("\n\r \t\u000b\u00a0\u2002"));
    }

```

The Unicode characters falling the tab (\t), seem to be incorrectly formatted. They should be `\u{000b}\u{00a0}\u{2002}`. Otherwise, they are just text and do not count as multi-byte whitespace.

petemcfarlane:
This fixes #190, thank you @MattLewin for reporting

### DIFF
--- a/exercises/bob/bob_test.php
+++ b/exercises/bob/bob_test.php
@@ -186,7 +186,7 @@ class BobTest extends PHPUnit\Framework\TestCase
 //    {
 //        $this->markTestSkipped();
 //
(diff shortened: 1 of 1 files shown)


MattLewin:
For some reason, I thought I was unable to make a PR on this. I’ll know better if I find another issue. Thanks for fixing it.

PR #195, opened 2018-02-04
Generate missing exercise READMEs

kytrinyx:
The following exercises were missing READMEs:

- collatz-conjecture
- crypto-square
- flatten-array
- transpose

In preparation for [v2](https://github.com/exercism/v2-feedback/blob/master/README.md) we're updating Configlet—our Exercism track linter—to require that we generate the exercise README. This is because in v2 we will not be generating READMEs on the fly, but deliver the README as defined in the directory for each implementation.

I'm going to go ahead and merge this as soon as the build passes, as this is janitorial work rather than a language-specific change.

See https://github.com/exercism/configlet/issues/116 for more details.

### DIFF
--- a/exercises/collatz-conjecture/README.md
+++ b/exercises/collatz-conjecture/README.md
@@ -1,0 +1,48 @@
+# Collatz Conjecture
+
+The Collatz Conjecture or 3x+1 problem can be summarized as follows:
+
+Take any positive integer n. If n is even, divide n by 2 to get n / 2. If n is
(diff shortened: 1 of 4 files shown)


kytrinyx:
We have a version conflict on Travis—the phptest.phar that is getting downloaded is too new for the version of PHP.

kytrinyx:
I'm actually going to merge this even though the build is passing, since the build is a separate problem. I'm submitting a PR with a suggested fix for the build.

PR #207, opened 2018-06-19
Add auto_approve to hello-world exercise

jackhughesweb:
This adds the auto_approve property to the hello-world 
exercise as part of preparing tracks for the v2 launch.

The current code on the server will automatically approve
hello-world exercises anyway, but that is a temporary fix
and will be removed in the future.

See exercism/php#206


kytrinyx:
Sweet, thanks!

PR #213, opened 2018-08-06
Remove markTestSkipped

### LINKED ISSUE #212: PHPUnit Groups
Hi,

We could start using PHPUnit Groups instead of `$this->markTestSkipped();`

# What would this fix

We don't have to remove `$this->markTestSkipped();` in order to run all tests.

# Example

```
/**
 * @group ignore
 */
public testSomeFunctionality()
{
}
```

To run tests that aren't in group `ignore`:

`phpunit --exclude-group=ignore` 

To run only tests that are in group `ignore`:

`phpunit --group=ignore`

Alternative would be adding to `phpunit.xml`:

```
<groups>
    <exclude>
        <group>ignore</group>
    </exclude>
</groups>
``` 

Which would allow to run just `phpunit` and only tests that don't have `ignore` groupe would run by default.

kunicmarko20:
closes #212 

Everything was search & replace so a review is needed!

### DIFF
--- a/exercises/accumulate/accumulate_test.php
+++ b/exercises/accumulate/accumulate_test.php
@@ -15,8 +15,6 @@ class AccumulateTest extends PHPUnit\Framework\TestCase
 
     public function testAccumulateSquares()
     {
-        $this->markTestSkipped();
-
         $accumulator = function ($value) {
             return $value ** 2;
         };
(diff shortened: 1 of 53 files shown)


kunicmarko20:
Looks fine to me, few extra lines were trimmed in the process. :smile:

PR #217, opened 2018-10-03
Adds the affine cipher exercise

camilopayan:
Needless to say, not finished yet! Just wanted to claim the exercise.

### DIFF
--- a/config.json
+++ b/config.json
@@ -658,6 +658,18 @@
         "lists",
         "text_formatting"
       ]
+    },
+    {
+      "slug": "affine-cipher",
+      "uuid": "d1267415-aff5-411d-b267-49a4a2c8fda2",
+      "core": false,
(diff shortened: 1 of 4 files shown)


camilopayan:
All right, this PR is ready for review! I'm going to go ahead and do another problem.

camilopayan:
Fixed the style issues! Thanks for the heads up on the `make style-check`
command.

petemcfarlane:
Think I've worked out why the build is failing.

>The exercise 'affine-cipher' is being unlocked by a non-core exercise. Non-core exercises can only be unlocked by core exercises.

So in the config you say it's unlocked by "grains" but "grains" is not a core exercise. Could you change this to a core exercise instead? Thanks 🙂

camilopayan:
All right, set it to null, which matches the rest, looks like.

PR #218, opened 2018-10-07
Adds blurb for PHP track

fernandovbs:
This change adds a blurb for the PHP track.


petemcfarlane:
Thanks @fernandovbs 😄

PR #219, opened 2018-10-12
Diamond exercise for php.

camilopayan:


camilopayan:
Ready for merge

camilopayan:
Fixed the indentation issues.

camilopayan:
I see there's an issue with locking and unlocking but I don't really get what that means.

petemcfarlane:
Same issue here. You could change the `unlocked_by` in the config, or just set it to null perhaps?

camilopayan:
Updated the `unlocked_by` to null

PR #220, opened 2018-10-15
Palindrome Products exercise

camilopayan:


camilopayan:
This is ready for review but I'm holding off on the `config.json` changes until I understand what the issue is with the other PRs.

camilopayan:
All right, set it up with `unlocked_by` set to null.

exercises/palindrome-products/palindrome-products_test.php
```diff
@@ -0,0 +1,104 @@
+<?php
+
+require "palindrome-products.php";
+
+class PalindromeProductsTest extends PHPUnit\Framework\TestCase
+{
+
+    public function testFindsTheSmallestPalindromeFromSingleDigitFactors()
+    {
+        list($value, $factors) = smallest(1, 9);
```
petemcfarlane: Do you think we should use the PHP 7.1 http://php.net/manual/en/migration71.new-features.php#migration71.new-features.symmetric-array-destructuring over `list()`? I can see arguments for both, using `[$value, $factors] = smallest(1, 9)` is more concise and subjectively _"better"_ PHP. But using `list(...)` is still supported and in older versions too.

camilopayan: Ah! I didn't know about this feature, but yes we should move to the destructuring syntax since PHP 5.6 is going to unsupported by the end of the year. I'm using it extensively in #221 so I'll go through and change those.

camilopayan:
All the `list`s were changed to `[]`.

PR #223, opened 2018-11-02
Update assertion arguments to expected, actual order for series exercise

dstockto:
Changed test name to be more indicative of what is tested

The assertion arguments were in the wrong order which makes the error messages confusing. If I had a failing test I'd see a message similar to "Expected 'your answer' but got 'correct answer'". This will put that back in order. Additionally I change the class name of the test since it doesn't seem to be related to a sieve.

### DIFF
--- a/exercises/series/series_test.php
+++ b/exercises/series/series_test.php
@@ -1,45 +1,45 @@
 <?php
 require "series.php";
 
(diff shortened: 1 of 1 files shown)


petemcfarlane:
Thanks @dstockto!

PR #224, opened 2018-11-02
Update crypto square tests expected vs actual ordering

dstockto:
This will make failing tests much more clear in their error message.


petemcfarlane:
Thanks!

PR #228, opened 2019-02-04
Restricted use of gmp_hamdist.

marvinrabe:
I saw students using the `gmp_hamdist` function to solve this exercise. I would rather prefer that they implement the behaviour themselves.


petemcfarlane:
Thanks!

Issue #230, opened 2019-02-24
A fatal error when running Bob tests


abachi:
Hi,
As said in the title I face a fatal error when I run `bob_test.php`.

Error Message:
`Fatal error: Declaration of BobTest::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void in C:\laragon\www\Exercism\php\bob\bob_test.php on line 156`

PHP version: PHP 7.3.2 (cli) (built: Feb  5 2019 23:14:21)
PHPunit: phpunit-8.0.4.phar
OS: Windows 7

DerTee:
phpunit8 is not supported right now, see #229

petemcfarlane:
We should probably update the tests to support PHPUnit 8, the current version as of the time of writing

petemcfarlane:
@abachi this should be working again now, thanks to @DerTee. Please report if you have any further issues, thanks!

PR #231, opened 2019-02-26
fix PHPUnit8 compatibility issues (#230)

DerTee:
PHPUnit 8 needs void return types for setUp, tearDown, etc. but only setUp is used in the exercises.

There is one small compatibility problem left for PHPUnit 8 that I don't know how to fix, but I guess that's fine for now:
When executing /php/exercises/grade-school/grade-school_test.php you get this warning:
```
PHPUnit 8.0.4 by Sebastian Bergmann and contributors.

..W...                                                              6 / 6 (100%)

Time: 167 ms, Memory: 10.00MB

There was 1 warning:

1) GradeSchoolTest::testAddStudentsinSameGrade
The optional $canonicalize parameter of assertEquals() is deprecated and will be removed in PHPUnit 9. Refactor your test to use assertEqualsCanonicalizing() instead.

WARNINGS!
Tests: 6, Assertions: 10, Warnings: 1.
[Finished in 0.3s]
```

Refactoring to assertEqualsCanonicalizing() is not possible because in PHPUnit7 that method doesn't exist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 .DS_Store
 tmp
 *.phar
+.phpunit.result.cache
 bin/configlet
 bin/configlet.exe
--- a/exercises/allergies/allergies_test.php
+++ b/exercises/allergies/allergies_test.php
@@ -77,7 +77,7 @@ class AllergiesTest extends PHPUnit\Framework\TestCase
         ], $allergies->getList());
     }
 
-    public function testIsAllergicToEgssAndShellfish()
+    public function testIsAllergicToEggsAndShellfish()
     {
         $allergies = new Allergies(5);
 
--- a/exercises/allergies/allergies_test.php
+++ b/exercises/allergies/allergies_test.php
@@ -58,40 +58,40 @@ class AllergiesTest extends PHPUnit\Framework\TestCase
     {
         $allergies = new Allergies(248);
 
-        $this->assertEquals([
+        $this->assertEqualsCanonicalizing([
             new Allergen(Allergen::CATS),
             new Allergen(Allergen::CHOCOLATE),
             new Allergen(Allergen::POLLEN),
             new Allergen(Allergen::STRAWBERRIES),
             new Allergen(Allergen::TOMATOES),
-        ], $allergies->getList(), "\$canonicalize = true", $delta = 0.0, $maxDepth = 10, $canonicalize = true);
+        ], $allergies->getList());
     }
 
     public function testIsAllergicToEggsAndPeanuts()
     {
         $allergies = new Allergies(3);
 
-        $this->assertEquals([
+        $this->assertEqualsCanonicalizing([
             new Allergen(Allergen::EGGS),
             new Allergen(Allergen::PEANUTS),
-        ], $allergies->getList(), "\$canonicalize = true", $delta = 0.0, $maxDepth = 10, $canonicalize = true);
+        ], $allergies->getList());
     }
 
     public function testIsAllergicToEgssAndShellfish()
     {
         $allergies = new Allergies(5);
 
-        $this->assertEquals([
+        $this->assertEqualsCanonicalizing([
             new Allergen(Allergen::EGGS),
             new Allergen(Allergen::SHELLFISH),
-        ], $allergies->getList(), "\$canonicalize = true", $delta = 0.0, $maxDepth = 10, $canonicalize = true);
+        ], $allergies->getList());
     }
 
     public function testIgnoreNonAllergenScorePart()
     {
         $allergies = new Allergies(509);
 
-        $this->assertEquals([
+        $this->assertEqualsCanonicalizing([
             new Allergen(Allergen::CATS),
             new Allergen(Allergen::CHOCOLATE),
             new Allergen(Allergen::EGGS),
@@ -99,7 +99,7 @@ class AllergiesTest extends PHPUnit\Framework\TestCase
             new Allergen(Allergen::SHELLFISH),
             new Allergen(Allergen::STRAWBERRIES),
             new Allergen(Allergen::TOMATOES),
-        ], $allergies->getList(), "\$canonicalize = true", $delta = 0.0, $maxDepth = 10, $canonicalize = true);
+        ], $allergies->getList());
     }
 
     /**
--- a/exercises/bob/bob_test.php
+++ b/exercises/bob/bob_test.php
@@ -9,7 +9,7 @@ class BobTest extends PHPUnit\Framework\TestCase
      */
     private $bob;
 
-    public function setUp()
+    public function setUp(): void
     {
         $this->bob = new Bob;
     }
--- a/exercises/bowling/bowling_test.php
+++ b/exercises/bowling/bowling_test.php
@@ -11,7 +11,7 @@ class GameTest extends PHPUnit\Framework\TestCase
     /** @var Game */
     private $game;
 
-    public function setUp()
+    public function setUp(): void
     {
         $this->game = new Game();
     }
--- a/exercises/grade-school/grade-school_test.php
+++ b/exercises/grade-school/grade-school_test.php
@@ -7,7 +7,7 @@ class GradeSchoolTest extends TestCase
 {
     protected $school ;
 
-    protected function setUp()
+    protected function setUp(): void
     {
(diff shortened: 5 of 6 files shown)


petemcfarlane:
Hmm but do you think we need to support PHPUnit 7 also? Could we just revert this PR https://github.com/exercism/php/pull/229 ontop of your commits @DerTee?

DerTee:
@petemcfarlane That's perfectly fine with me if our focus is to primarily support the latest versions of PHP and PHPUnit.
A few PHP students might get errors if they try to execute grade-school_test.php with PHPUnit 7, but that's it. Not a big problem imho.

I'll do the revert of PR #229 and change to assertEqualsCanonicalizing later today.

DerTee:
Hmm, that's what I forgot: PHPUnit 8 breaks PHP 7.1 compatibility:

```
$ make test
running tests for: accumulate
PHPUnit 8.0.4 by Sebastian Bergmann and contributors.
This version of PHPUnit is supported on PHP 7.2 and PHP 7.3.
You are using PHP 7.1.11 (/home/travis/.phpenv/versions/7.1.11/bin/php).
make[1]: *** [test-assignment] Error 1
make: *** [test] Error 1
The command "make test" exited with 2.
```

So, what do you think? Go with PHPUnit 8 and lose PHP 7.1 support or stick to PHPUnit 7 and live with that one little deprecation warning in PHPUnit 8?

Or a little more general: Do we care more about using the most recent software versions or about backwards compatibility?

I'm fine with either. I care more about telling the students up front which versions they can expect to work.

petemcfarlane:
Hmm maybe we stick with PHPUnit 7 a little longer, but can we update the test anyhow to remove the deprecation warning? I think if we can support PHPUnit 7 and 8 at little extra cost for now, then that should work for _most_ peoples environments?

rossbearman:
For what it's worth, I've only been mentoring people for two days and I've had to explain to two people how to fix the failure over the missing `void` declaration on the `setUp`. However fixing that would remove suppport for PHPUnit 7/PHP 7.1.

It would seem to me that most students going forward are going to be using the latest version of PHP and PHPUnit, as they would be running it locally where upgrading is generally less of an issue. Given that  the instructions would need to be changed to tell people to download a previous release, would it not make more sense to support the latest stable release instead? Is it likely that many people would be stuck on PHP 7.1 on their local machine?

DerTee:
@petemcfarlane 
Updated my PR. I think it should be ok now, but let me know if something needs to be changed.
Sticking to PHPUnit 7 for now seems like a good idea to me. Also there's basically no extra work involved, except for using the newer assertions, which were backported to PHPUnit 7 anyway.

That said, switching to assertEqualsCanonicalize might puzzle some students (like me, I was using PHPUnit 7.4.4 till today), since that function was only introduced in PHPUnit 7.5 (see the [changelog](https://github.com/sebastianbergmann/phpunit/blob/7.5/ChangeLog-7.5.md)), which was released in December 2018.

@rossbearman 

> For what it's worth, I've only been mentoring people for two days and I've had to explain to two people how to fix the failure over the missing void declaration on the setUp. However fixing that would remove suppport for PHPUnit 7/PHP 7.1.

Are you sure adding return type declarations breaks compatibility? Those were added in PHP 7.0 if I'm reading the [docs correctly](https://secure.php.net/manual/en/functions.returning-values.php#functions.returning-values.type-declaration) (it's a little unclear though):

> PHP 7 adds support for return type declarations. [...] As of PHP 7.1.0, return values can be marked as nullable by prefixing the type name with a question mark (?). This signifies that the function returns either the specified type or NULL.

@G-Rath I just realized that I was making the pretty much same changes as you in PR #238

rossbearman:
>Are you sure adding return type declarations breaks compatibility? Those were added in PHP 7.0 if I'm reading the docs correctly (it's a little unclear though):

You're right, they won't break compatibilty. I was mistakenly thinking that adding the return type would break the `setUp()` contract in PHPUnit 7, not that they weren't present in PHP 7.0. But that isn't the case, I've just tested it and it works fine.

rossbearman:
>That said, switching to assertEqualsCanonicalize might puzzle some students (like me, I was using PHPUnit 7.4.4 till today), since that function was only introduced in PHPUnit 7.5 (see the changelog), which was released in December 2018.

PHPUnit does support the `@requires` docblock directive, however that just makes it skip the test if the version requirement isn't met, you'd still have to look in the file to see why it was skipped. Perhaps there is a way to cause that to raise an error? Alternatively it could just go in the docs.

```
/**
 * @requires PHPUnit 7.5.0
 */
```

DerTee:
@rossbearman Thanks, didn't know about the `@requires` directive! 
I just tried using it to have two different tests, one for PHPUnit 7.5 and up and for those below, but unfortunately that doesn't work. `@requires` seems to only care whether the current version is higher.

I think the error message as you would get right now with PHPUnit 7.4 is ugly but fairly straightforward for students and mentors:
```
PHPUnit 7.4.4 by Sebastian Bergmann and contributors.

.................EEEE........                                     29 / 29 (100%)

Time: 179 ms, Memory: 10.00MB

There were 4 errors:

1) AllergiesTest::testIsAllergicToLotsOfStuffs
Error: Call to undefined method AllergiesTest::assertEqualsCanonicalizing()

D:\tmp\exercism-repos\php\exercises\allergies\allergies_test.php:61

[...]

ERRORS!
Tests: 29, Assertions: 81, Errors: 4.
[Finished in 0.3s]
```

Skipped tests might be more easily ignored by students, so they would upload their solution without saying anything to the mentor about it.

da-moo:
Student here. I started the PHP track not too long ago and just ran into this issue on the 'bob' exercise. I agree with @rossbearman in that I don't think it makes much sense to ask students to downgrade their version of PHP or PHPUnit to run tests. Unfortunately I'll have to stop my progress on the PHP track until this PR and corresponding issue(s) are resolved. Best of luck

DerTee:
@da-moo Actually you can continue your exercise if you want. You can just download the [modified test file 
 for bob](https://github.com/exercism/php/raw/2f15393b90080e99f945399b392b785ce97266ed/exercises/bob/bob_test.php) and use that or add the return type in the test file yourself.

da-moo:
> @da-moo Actually you can continue your exercise if you want. You can just download the [modified test file for bob](https://github.com/exercism/php/raw/2f15393b90080e99f945399b392b785ce97266ed/exercises/bob/bob_test.php) and use that or add the return type in the test file yourself.

I appreciate the info but I'd rather not have to modify tests by either copying it down from a branch or modifying it directly every time I want to do a new exercise. I'm okay with waiting until the PHP track gets updated.

rossbearman:
Hi Peter,

It is only a couple of exercises that require modification and it's just a case of adding `: void` to the `setUp()` method signature in the test file that Exercism downloads.

da-moo:
> 
> It is only a couple of exercises that require modification and it's just a case of adding `: void` to the `setUp()` method signature in the test file that Exercism downloads.

Ah I didn't realize that's all that was involved to make the tests PHPUnit 8 compatible. I must've misunderstood the issue and PR, my mistake. I don't mind adding a type hint to one method if that's all that's required. Thanks Ross.

rossbearman:
The other issue that's being fixed by the PRs just throws a deprecation warning in PHPUnit 8.

petemcfarlane:
Ok, sorry for the delay in merging this. I think we're all agreed this is the best solution (works for largest majority of uses) at this time. If the missing `assertEqualsCanonicalizing` is causing a lot of problems for people, then we could not use it at all and instead just call `sort` on the resulting array - or make that part of the exercise...

James-N-M:
Hey guys I'm on the grade-school challenge and am getting this error. I read through this prs comments and am not seeing what solution was decided upon lol. php unit 7.2.7 and php 7.2.10 :)
Add studentsin same grade
   │
   │ Error: Call to undefined method GradeSchoolTest::assertEqualsCanonicalizing()
   │
   │ /Users/jamesm/Exercism/php/grade-school/grade-school_test.php:31

DerTee:
@James-N-M 
You are right, there is now a wall of text but the final solution wasn't stated clearly, so here it is:
**Tests now require  PHPunit 7.5 or later and PHP 7.1 or later**

James-N-M:
Thank you !

PR #234, opened 2019-02-28
Replace usage of deprecated phpunit expected exception tag

### LINKED ISSUE #232: Replace usage of deprecated `@expectedException` with `expectException`
Some exercises are using the `@expectedException` annotation, which is deprecated and will be removed in PHPUnit 9:

```
The @expectedException, @expectedExceptionCode, @expectedExceptionMessage, and @expectedExceptionMessageRegExp annotations are deprecated. 

They will be removed in PHPUnit 9. Refactor your test to use expectException(), expectExceptionCode(), expectExceptionMessage(), or expectExceptionMessageRegExp() instead.
```

The affected exercises should be updated to use `expectExceptionMethod`.

(I'm happy to do this myself, when I've got some free time 😄)

The following exercises require updating:
   - [ ] - `largest-series-product`
   - [ ] - `phone-number`
   - [ ] - `ocr-numbers`
   - [ ] - `triangle` 
   - [ ] - `grains`
   - [ ] - `robot-simulator`
   - [ ] - `queen-attack` 

(This is a 'living' list - I'll add exercises as I found them)

G-Rath:
This closes #232.

While doing this I noticed usage of `expectException` like this:

`$this->expectException('InvalidArgumentException', 'Only positive numbers are allowed');`

Personally, I think it'd be better to use `InvalidArgumentException::class`, but more importantly this doesn't actually test the exception message; PHPUnit just ignores the second parameter, checking only the exceptions *type*.

To achieve the desired behaviour, `expectExceptionMessage` should be used. 

I'll open up a new issue for this :)

### DIFF
--- a/exercises/grains/grains_test.php
+++ b/exercises/grains/grains_test.php
@@ -51,9 +51,6 @@ class GrainsTest extends PHPUnit\Framework\TestCase
         $this->assertSame('9223372036854775808', square(64));
     }
 
-    /**
-     *
-     */
     public function testRejectsZero()
     {
         $this->expectException(InvalidArgumentException::class);
@@ -61,9 +58,6 @@ class GrainsTest extends PHPUnit\Framework\TestCase
         square(0);
     }
 
-    /**
-     *
-     */
     public function testRejectsNegative()
     {
         $this->expectException(InvalidArgumentException::class);
@@ -71,9 +65,6 @@ class GrainsTest extends PHPUnit\Framework\TestCase
         square(-1);
     }
 
-    /**
-     *
-     */
     public function testRejectsGreaterThan64()
     {
         $this->expectException(InvalidArgumentException::class);
--- a/exercises/grains/grains_test.php
+++ b/exercises/grains/grains_test.php
@@ -52,26 +52,32 @@ class GrainsTest extends PHPUnit\Framework\TestCase
     }
 
     /**
-     * @expectedException InvalidArgumentException
+     *
      */
     public function testRejectsZero()
     {
+        $this->expectException(InvalidArgumentException::class);
+
         square(0);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


DerTee:
@G-Rath 
Thanks for doing all that maintenance work and updating the tests!

Just out of curiosity:
What's the benefit of having one commit per file in this case?
Personally I try to squash commits that are related to the same conceptional change in one commit so the log stays lean.
Am I missing something?

G-Rath:
It's considered a recommended [best practice](http://sethrobertson.github.io/GitBestPractices/#commit) for git: "Commit Often, Perfect Later".

> Git only takes full responsibility for your data when you commit. If you fail to commit and then do something poorly thought out, you can run into trouble. Additionally, having periodic checkpoints means that you can understand how you broke something.

Lean commits are great until you want to do something related to the changes in them.

For example, in this PR, pretend all the tests suddenly broken. We could checkout each commit at a time (usually I'd pick a random commit every couple of commits until I found one that didn't error, then checkout one at a time from that commit up until my tests started failing, rather than do them strictly one-by-one going down), and run the tests - you'd do that until they failed, then you'd know what code change caused the failure.

However, if you'd only made one big commit, you'd have to *manually* revert a part of that single commit to test against, and keep doing that until you'd found the failure.

You can still of course do this with a single commit, but it's a lot easier to let git manage it all for you :)

petemcfarlane:
Thanks for the PR @G-Rath, much better! But don't you think the empty docblocks above the tests look a little strange and potentially confusing now? Would you consider removing these too? Thanks!

G-Rath:
@petemcfarlane Sorry, I thought I replied to your comment, but apparently I didn't.

I totally agree it's weird! Also, sorry for forgetting to add "FIXES" to all my PRs :(

PR #239, opened 2019-03-06
Add "number of students" test to grade-school

rossbearman:
Adds a missing test for the `numberOfStudents()` method which was called from the existing tests, but only to check whether there are no students listed before any are added. The new test adds five students to different grades and then checks the method returns `5`.

This requirement should possibly also be added to the README as currently it's only discoverable through the tests. Alternatively the existing test `testNoStudent()` could be removed to remove this requirement.

### DIFF
--- a/exercises/grade-school/grade-school_test.php
+++ b/exercises/grade-school/grade-school_test.php
@@ -57,6 +57,17 @@ class GradeSchoolTest extends TestCase
         $this->assertEmpty($this->school->grade(1));
     }
 
+    public function testNumberOfStudents()
+    {
+        $this->school->add("Marc", 1);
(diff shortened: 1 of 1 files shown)


rossbearman:
Having looked at a few other language's implementations of these tests, it seems like PHP is the only one that has a test to check that there are no students at the beginning, so it might just make more sense to remove the `testNoStudent()` method as it's not part of the exercise description.

petemcfarlane:
Yeah, I'm all for removing that `testNoStudent` case, it doesn't add a lot of value IMO and makes the exercise simpler, thanks

rossbearman:
Great, I'll close this and open a PR to remove that test.

PR #240, opened 2019-03-10
Correct mention of non-ASCII symbols in Pangram

rossbearman:
The README for the Pangram problem states that the input will not contain non-ASCII symbols, however the test `testPangramWithNonAsciiCharacters()` does, as the name suggests, include non-ASCII characters. This updates the README to be consistent with the tests.


petemcfarlane:
Whoops, good catch, thanks!

PR #242, opened 2019-03-11
Move Run Length Encoding later in the track order

rossbearman:
Place the Run Length Encoding exercise between "Isogram" and "Difference of Squares", giving it a more appropriate location for it's difficulty level and increase it's difficulty level to 5, to reflect the complexity of solutions that are seen for it and bring it in line with "Largest Series Product".

For comparison, the C# track gives it a difficulty rating of 5 and the Ruby track assigns this a difficulty of 4.

### DIFF
--- a/config.json
+++ b/config.json
@@ -48,18 +48,6 @@
       ]
     },
     {
-      "slug": "run-length-encoding",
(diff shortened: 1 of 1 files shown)


rossbearman:
Superceded by #246.

PR #243, opened 2019-03-11
Move Raindrops earlier in the track order.

rossbearman:
Place the Raindrops exercise between "Gigasecond" and "Bob", giving it a more appropriate location for it's difficulty level and bringing it in line with other language tracks that place it shortly after "Hamming". This also reduces it's difficulty rating from 6 to 2, bringing it inline with "Hamming" and "Gigasecond".

### DIFF
--- a/config.json
+++ b/config.json
@@ -37,6 +37,17 @@
       ]
     },
     {
+      "slug": "raindrops",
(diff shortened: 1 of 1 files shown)


rossbearman:
Superceded by #246.

Issue #244, opened 2019-03-11
Formalising the PHP core track


rossbearman:
While opening pull requests to reorder Raindrops and Run Length Encoding I noticed that the PHP track doesn't actually have a specified core track, Exercism is just automatically pulling the first eleven exercises from the `config.json` file. As a result, there are a lot of string manipulation exercises and not a whole lot else until the last two.

To start some discussion on the order I've put together the following track proposal based on the kinds of solutions that I've seen for the existing problems, the concepts introduced by each problem as well as the track ordering for JavaScript, Ruby, Python, Go and Elixir.

1. Hamming
2. Gigasecond
3. Bob
4. RNA Transcription
5. Luhn
6. Isogram
7. Robot Name
8. Difference of Squares
9. Grade School
10. Robot Simulator
11. Run Length Encoding
12. Largest Series Product

## Notes

- Removed Pangram as it doesn't usefully introduce anything over the other string manipulation exercises and Isogram is present in most other tracks.
- Removed Raindrops due to it's similarity to Bob and added Luhn as a candidate to replace it.
- Moved Run Length Encoding to the penultimate position, due to the amount of mentoring it can require and propensity to receive spaghetti solutions.
- Moved Difference of Squares earlier in the chain, as it's a simpler exercise.
- Added Robot Name as it introduces randomness and can be used to teach about separation of concerns.
- Added Grade School as it introduces sorting and can be solved with a number of different approaches to the data structure.
- Added Robot Simulator as it introduces multi-dimensional arrays and more complex class structuring.

## Further considerations
- I'm not sure about the position of Run Length Encoding. My thinking with placing it in the penultimate position is that it's a lot easier to mentor if the student doesn't need much help with code style and clarity, only the concepts.
- I'm not surre about the position of Difference of Squares, this could possibly come even earlier?
- There are probably a couple of other side exercises that could be moved into the core track to widen the concepts covered.
- The order could be further improved, I'm particularly unsure of where RNA Transcription and Isogram should fit.
- Is Largest Series Product the best final exercise? It would be good to have the final exercise incorporate a number of the concepts introduced earlier, especially focusing on class structure, array and string manipulation. I'm not sure what exercise would be best provide this, however.

dstockto:
I'm not so sure about removal of run-length-encoding. It's a good introduction to preg_replace_callback. While it may not be obvious, and it's probably not unless someone knew that function already, I think it's a good intro to that and some of the huge number of purpose-built functions PHP has.

[...]

PR #246, opened 2019-03-12
Implement a defined core track and reorder the default exercises

rossbearman:
This PR implements the changes proposed in #244. This is only intended as a first step and could certainly do with further refinement, but is hopefully an improvement on the current progression of the core track and is more in line with the goals of the Track Anatomy Project.


rossbearman:
Closing this to rebase on top of the latest changes and modify the order.

PR #247, opened 2019-03-13
Implement a defined core track and reorder the default exercises

rossbearman:
This PR implements the changes proposed in #244. This is only intended as a first step and could certainly do with further refinement, but is hopefully an improvement on the current progression of the core track and is more in line with the goals of the Track Anatomy Project.

The new core track would be as follows:

1. Hamming
2. Gigasecond
3. Bob
4. RNA Transcription
5. Luhn
6. Isogram
7. Robot Name
8. Difference of Squares
9. Grade School
10. Robot Simulator
11. Run Length Encoding
12. Largest Series Product

### DIFF
--- a/config.json
+++ b/config.json
@@ -6,7 +6,7 @@
     {
       "slug": "hello-world",
       "uuid": "56116aa6-e3d1-4122-a394-a7e0f6ab6fea",
-      "core": false,
+      "core": true,
       "auto_approve": true,
       "unlocked_by": null,
       "difficulty": 1,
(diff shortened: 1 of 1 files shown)


petemcfarlane:
Thanks for working on this @rossbearman and @dstockto, I'll merge this as it is and we can refine in future if we find any more ideas!

rossbearman:
Thanks @petemcfarlane! I avoided setting any of the side exercises to have requirements for unlocking for the moment, but that might be something to think about.

PR #248, opened 2019-03-13
Remove testNoStudents() and clean up Grade School example

rossbearman:
This PR removes the testNoStudents() method from Grade School. Additionally, the second commit cleans up the Grade School example, which had some weird formatting (spaces before semi-colons and around array accessors) and makes the sorting method clearer.

Let me know if this PR would be preferred without the example change, I just thought I'd do it while making the test change.

### DIFF
--- a/exercises/grade-school/example.php
+++ b/exercises/grade-school/example.php
@@ -2,29 +2,26 @@
 
 class School
 {
-    private $database = array() ;
+    private $students = [];
 
-    public function add($student, $grade)
+    public function add($name, $grade)
     {
-        $this->database[$grade][] = $student ;
+        $this->students[$grade][] = $name;
     }
 
     public function grade($grade)
     {
-        return (array_key_exists($grade, $this->database) ? $this->database[$grade] : array()) ;
+        return (array_key_exists($grade, $this->students) ? $this->students[$grade] : []);
     }
 
     public function studentsByGradeAlphabetical()
     {
-        $tmp = $this->database ;
-        ksort($tmp) ;
+        ksort($this->students);
 
-        foreach ($tmp as $grade => $students) {
-            asort($students) ;
-            foreach ($students as $student) {
-                $res [ $grade ][] = $student ;
-            }
-        }
-        return $res ;
+        return array_map(function ($grade) {
+            sort($grade);
+
+            return $grade;
+        }, $this->students);
     }
 }
--- a/exercises/grade-school/example.php
+++ b/exercises/grade-school/example.php
@@ -4,11 +4,6 @@ class School
 {
     private $database = array() ;
 
-    public function numberOfStudents()
-    {
-        return (count($this->database, COUNT_RECURSIVE) - count($this->database)) ;
-    }
-
     public function add($student, $grade)
     {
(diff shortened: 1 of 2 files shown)


exercises/grade-school/example.php
```diff
@@ -2,29 +2,26 @@
...
     {
-        return (array_key_exists($grade, $this->database) ? $this->database[$grade] : array()) ;
+        return (array_key_exists($grade, $this->students) ? $this->students[$grade] : []);
     }
 
     public function studentsByGradeAlphabetical()
     {
-        $tmp = $this->database ;
-        ksort($tmp) ;
+        ksort($this->students);
```
petemcfarlane: I'm not even sure why this is necessary, but it was already here so 🤷‍♂️

rossbearman: Hmm, there's something a bit weird going on here.

`ksort($this->students)` should be required because the order of the grades themselves matters. `testSortSchool()` adds several students to grade 5 and then one to grade 4, which results in `$this->students` looking like this:

```
    5 => ['Claire', 'Marc', 'Virginie'],
    4 => ['Mehdi'],
```

But `testSortSchool()` is expecting this (sorted by grade, and student name within a grade):

```
    4 => ['Mehdi'],
    5 => ['Claire', 'Marc', 'Virginie'],
```

However, removing the sort doesn't break the test (and it's definitely returning it out of order). The test is using `assertEquals()`, not `assertEqualsCanonicalizing()`, so it should fail. Replacing the assertion with `assertTrue($sortedStudents === $schoolStudents)` _does_ make it fail.

I assume I'm missing something obvious, but I'm not sure what that is.

petemcfarlane: The keys are 4 and 5 both time though, I think that is enough for testing with `assertEquals`?

rossbearman: I thought that was supposed to be the purpose of assertEqualsCanonicalizing(), as it's only difference is that it sorts the array before performing the comparison?

Either way, the tests should probably be changed, as the requirement from the is that the array is sorted by grade: "Grades should sort as 1, 2, 3, etc., and students within a grade should be sorted alphabetically by name. "

rossbearman: I just looked into the PHPUnit source, and `assertEqualsCanonicalizing()` or the `$canonicalize` parameter uses `sort()` on both the expected and actual arrays, which strips out the keys. I had been under the impression it was using `ksort()` or `asort()`.

And you're correct that `assertEquals()` on it's own simply iterates over the expected array, checking the for existence of a key and that it's value matches.

The test should probably be changed to:

`assertTrue($sortedStudents === $schoolStudents)`

petemcfarlane: Hmm yeah I suppose that would be the most correct. I probably wouldn't have bothered to check that the keys are sorted, only it is stated in the README, so perhaps we should in this case?

rossbearman: Opened #249.

I think it makes sense to check, as the idea is presumably that in a real world scenario you would be displaying the students in each grade, ordered by their grade.

PR #250, opened 2019-03-17
Improve Hamming skeleton variables and method signature

rossbearman:
As this is supposed to be a starting point for the exercise and this is the first real exercise that a student will carry out, it seems like the provided code should follow best practices. This PR introduces four changes:

1) Replace `$a` and `$b` with `$strandA` and `$strandB`.
2) Introduce type hinting and return type declaration.
3) Replace multiple `//` comments at the top with a multiline comment.
4) Suggest that the student removes the skeleton file comment, as this gets left in the vast majority of solutions that are submitted.

### DIFF
--- a/exercises/hamming/hamming.php
+++ b/exercises/hamming/hamming.php
@@ -1,11 +1,13 @@
 <?php
 
-//
-// This is only a SKELETON file for the "Hamming" exercise. It's been provided as a
-// convenience to get you started writing code faster.
-//
+/*
+This is only a SKELETON file for the "Hamming" exercise. It's been provided as a
+convenience to get you started writing code faster.
 
-function distance($a, $b)
+Remove this comment before submitting your exercise.
+*/
+
+function distance(string $strandA, string $strandB): int
 {
     //
     // YOUR CODE GOES HERE


dstockto:
I'm a bit on the fence on most of the changes. I am definitely on board with changing the variable names to something better than `$a` and `$b`. 

I'm a bit torn on the type-hinting and return types. I am 100% for them in my own code and I tell students that I like that they used them when they submit using them, but it's also not something that's required by PHP and it's also fairly new and somewhat controversial across PHP devs. 

For the comment change, I've rarely seen and rarely ever use /* ... */ outside of a function body. For doc blocks I'd use /** ... */ but this is not a doc block situation. When I see /* ... */ it's usually inside a function to explain a particularly tricky or complicated bit of code. So I could go either way on this one. If you feel strongly about changing the comment style then that's fine.

As far as adding the comment about removing the comment, I like that. I would guess 60-80% come in with the comment still there. I thought I'd seen it somewhere to remove it (maybe another track) or it could have been a mentor that mentioned it was no longer valid. Since then, I do the same when I see these.

rossbearman:
I admittedly opened this pull request solely to improve the variable names, I decided to add the other changes when I was already in the file. I wasn't sure if the type hints would be well received, but decided that as the types in this instance are clear and we're not enforcing strict type checks there is no real downside to them, but it does expose students to their existence. It's also a good habit to get into when it comes to using automated code tools and static analysis. I can see the argument against including them however, as they're not required and some students may not realise that. 

I'm happy either way with the comments, you'd never normally see non-docblock comments outside of a function body or class, so it's an exceptional case, but generally I would prefer to see multiline comments _inside_ a function using the multiline comment syntax, so it seemed better to present that way of doing it. No strong feeling either way.

dstockto:
I think it's overall a good change. I would not have an issue if it were merged.

petemcfarlane:
I think it's a useful way to introduce the student to the idea of type hinting and return types, so I would be in favour of this change. If anyone has any reason to disagree, we can always reopen a new issue and debate further!

PR #251, opened 2019-03-18
Assign unique uuids

F3PiX:
Some exercises have their UUID copied from other tracks. That causes side effects in the _other_ track.
This PR adds new, unique UUID's to solve that.

@ihid 

DON"T MERGE PLEASE, wait for @ihid to merge.


iHiD:
Thank you.

PR #255, opened 2019-05-23
Add test to stop non-numerics passing when casting is used

dstockto:
Several solutions have come across that pass only because they
cast each individual character to an integer. This commit adds
a test for an input that is matches the five zeros input, except
the characters are non-numeric. This should be invalid, but
solutions that cast character to int will allow this one which is
not correct.


dstockto:
The linting issue should be fixed. I was using an EAP of my IDE and it seems to have a formatting issue currently.

dstockto:
This should now be actually fixed. The checks seem to say that too. Sorry about the previous push.

PR #257, opened 2019-06-12
Cleaning up code to comply with PSR12

yisraeldov:
There are still some problems with namespaces

### DIFF
--- a/exercises/acronym/example.php
+++ b/exercises/acronym/example.php
@@ -15,6 +15,6 @@ function acronym($phrase)
     }
 
     return array_reduce($words[0], function ($acronym, $word) {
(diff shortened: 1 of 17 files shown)


petemcfarlane:
Hey thanks for the PR.  My understanding is PSR12 is still under review, do you think it makes sense to wait for it to be finalised before committing this, just in case there are any breaking changes?

yisraeldov:
Well it is really trivial to run phpcbf and fix it again, but I don't see
that it would be any breaking changes, mostly just adding spaces around the
`.` operator.

I think we should be encouraging people to use PSR12 since it is compatible
with PSR2 and it takes into account the differences in PHP that where made
in the last 10 years (or so )

PR #262, opened 2019-07-02
Add class constant modifiers

arueckauer:
PHP supports class constants visibility modifiers as of PHP 7.1.0, see https://www.php.net/manual/en/language.oop5.constants.php


petemcfarlane:
Fair enough, PHP versions earlier than 7.1 are out of active support already

PR #264, opened 2019-07-02
Replace class name with self for self-references

arueckauer:


petemcfarlane:
Is there any commentary or examples online for why using `self` is preferred over the other classname?

arueckauer:
Good question. Right off the top of my head I cannot think of any documentation.

Maybe it is a personal preference. I like to use it, because (1) it explicitly shows that the reference is for the current class and (2) makes refactoring very easy.

Right now the codebase has both cases. Personally I would prefer a constistent usage.

petemcfarlane:
Yeah fair enough. I tend to use `self` and `static` from inside a class, but just wondered if there were any other benefits other than those you mention.

PR #268, opened 2019-07-03
Define indentation of 4 spaces for PHP files

arueckauer:
> Code MUST use 4 spaces for indenting, not tabs.

See https://www.php-fig.org/psr/psr-2/


arueckauer:
`indent_style` is inherited from `[*]` and thus applied to `[*.php]`.

You want to explicitly define it?

petemcfarlane:
Ah sorry, in the PR GitHub only showed me the surrounding few lines, I didn't see the whole file 🤦‍♂ I see it is defined already, as you say so 👍

PR #272, opened 2019-07-03
Convert unnecessary double quotes to single quotes

arueckauer:
Unify code style


arueckauer:
Good catch! Fixed both.^^

arueckauer:
I close this, since resolving merge conflicts will be a much larger task than the actual changes provided.

Issue #260, opened 2019-04-10
PHP Track - Phone Number Exercise


mmoray:
The unit tests for the exercise are not built correctly according to the README.md specifications. The N is supposed to be a number between 2 - 9 and the unit tests do not test this and furthermore allow for 0 and 1 for the N number.

petemcfarlane:
Hi @mmoray thanks for raising the issue. You can see the canonical data test file here from the exercism problem specification repository https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json

Would you care to create a PR to fix this issue? If not perhaps someone else might like to pick this issue up?

arueckauer:
@petemcfarlane I believe, this can be closed now.

PR #274, opened 2019-07-03
Fix tests and example of phone number

arueckauer:
This PR resolves issues discussed in #260:

Adding tests for
- example inputs in readme
- validating first numbers of area code and prefix (allowing 2 through 9)

Resolve conflicts in existing tests due to added validation of area code and prefix.

### DIFF
--- a/exercises/phone-number/example.php
+++ b/exercises/phone-number/example.php
@@ -19,27 +19,17 @@ class PhoneNumber
         return substr($this->number, 0, 3);
     }
 
-    protected function prefix() : string
-    {
-        return substr($this->number, 3, 3);
-    }
-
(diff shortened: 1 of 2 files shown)


arueckauer:
@petemcfarlane Thanks for pointing that out. Did not know about `canonical-data.json`. Will fix!

arueckauer:
Tests are fixed as discussed.

One point that is still missing, are the assertions of expected Exception messages. I recommend opening another issue for that / providing that in another PR, since this requires updating the example as well.

arueckauer:
Just had the time to add assertions of exception messages.

petemcfarlane:
Super thanks!

PR #276, opened 2019-07-17
Improve Gigasecond exercise example

arueckauer:
This PR improves the example of the Gigasecond example.

- Remove unnecessary clone of `DateTimeImmutable` object
- Declare return type

### DIFF
--- a/exercises/gigasecond/example.php
+++ b/exercises/gigasecond/example.php
@@ -3,6 +3,5 @@
 function from(DateTimeImmutable $from)
 {
     $interval = new DateInterval('PT1000000000S');
-    $date = clone $from;
-    return $date->add($interval);
+    return $from->add($interval);
 }
--- a/exercises/gigasecond/example.php
+++ b/exercises/gigasecond/example.php
@@ -1,6 +1,6 @@
 <?php
 
-function from(DateTimeImmutable $from)
+function from(DateTimeImmutable $from) : DateTimeImmutable
 {
     $interval = new DateInterval('PT1000000000S');
     return $from->add($interval);


dstockto:
Changing to use DateTimeImmutable means the no mutation test doesn't really have a purpose any more.

arueckauer:
Hey David, thanks for pointing that out. What are you suggesting, removing the no mutation test? 
As reference, `DateTimeImmutable` was introduced in #267 .

dstockto:
@arueckauer Potentially. The change to DateTimeImmutable makes the solution a one-liner. Previously it was a way I could introduce the concept of cloning objects. In the past several exercises I've mentored, most were one-liners. One introduced a function that had a type declaration of DateTimeInterface which is a pretty cruddy option, but it does mean that clone would be needed unless the code actually determined and branched on whether it was a DateTime or a DateTimeImmutable, or it could use just the methods in DateTimeInterface (which suck).

If we include a sample starter file with the type declaration of DateTimeImmutable then I think the clone test is definitely not needed and could be removed.

If we go with a starter of DateTimeInterface, then clone should remain and we can mentor this with clone and the cruddy API that that interface provides. If we go with this case, we should also probably add some tests that use DateTime in addition to DateTimeImmutable.

arueckauer:
@dstockto I see. The original intention was to discourage the usage of `DateTime` since it can lead to unexprected behaviour. But I agree, `clone` is a valueable lesson here.

Right now I am in favor of your last suggestion, adding type declarations for `DateTimeInterface` and tests for both implementations. However I am reluctant, since I cannot imagine clean code that without having to differ depending provided object type. But maybe it's not that bad (?).

Testing two different functions, each accepting either `DateTime` or `DateTimeImmutable` would probable be too much.

If that is okay with you, I can look into the addressed solution.

dstockto:
I like the discouraging of using DateTime. I wonder if there's a new way or different exercise we could introduce to bring up clone. 

As for using DateTimeInterface, I think that's nearly as bad as DateTime. It's a bad interface that doesn't seem to cover the good parts of the DateTimeImmutable or DateTime classes. This will lead to probably some pretty dumb code overall (I'd guess?). Either you have to have a conditional to determine if you're dealing with DateTime or DateTimeImmutable and then do different things or you have to use just the minimal stuff that the interface provides which leads to a pattern where you have to extract the time info from the incoming parameter and make a new object with that value plus the billion seconds or ... I don't know.

I suppose I'm leaning towards accepting and returning DateTimeImmutable, removing the clone test and then perhaps seeing if there's a different exercise that clone could be used. Or perhaps we don't need to worry about it. Outside of exercism's gigasecond problem, I rarely use clone.

petemcfarlane:
Hey, I guess you could implement the [Clock](https://github.com/exercism/php/blob/f2cf7d7ea9d94c6ab16efbe6d7f1966c4b24002e/exercises/clock/example.php#L34) exercise with `clone`? Though I realise now that there isn't actually a test to check that the `Clock`  is immutable, though the tests are written an that style, e.g. `$clock = $clock->add(...)` instead of `$clock->add(...)`. 
What do you think about adding an immutable test for Clock, and sticking to `DateTimeImmutable` here?

arueckauer:
@dstockto Good points. Right now tests are open to usage of either `DateTime` or `DateTimeImmutable` instances. We could stricten tests to only allow instances of `DateTimeImmutable`. What do you think?

dstockto:
I think that's fine to make the tests (and potentially the starter file) to just use DateTimeImmutable.

PR #284, opened 2019-08-05
Create stub files for all exercises

arueckauer:
This PR provides changes requested in #282.

Exisiting stub files are left untouched

- exercises/accumulate/accumulate.php
- exercises/binary/binary.php
- exercises/hamming/hamming.php
- exercises/hello-world/hello-world.php
- exercises/markdown/markdown.php
- exercises/run-length-encoding/run-length-encoding.php


petemcfarlane:
This is a good idea IMO, but I think we should keep the markdown exercise, as that is a slightly different problem:

> The markdown exercise is a refactoring exercise. There is code that parses a given string with Markdown syntax and returns the associated HTML for that string. Even though this code is confusingly written and hard to follow, somehow it works and all the tests are passing! Your challenge is to re-write this code to make it easier to read and maintain while still making sure that all the tests keep passing.

Do you think we can keep the markdown stub as it is?

arueckauer:
Oh, the second commit accidently overwrote the existing stubs. Sorry for that, I'll fix that tomorrow (afk now).

arueckauer:
I will provide changes in a new PR, to not mess up history of existing stubs files.

Issue #81, opened 2016-05-26
Anagram solutions require libraries not included in some default PHP installs.


lafent:
Solutions for the anagram exercise would typically take advantage of mb_string function.  Some (default) PHP installs won't include this extension, which could lead a user to being unable to complete the exercise if they are unable or unwilling to install the extension.

Can we move the tests that use UTF-8 based characters to being a stretch goal, similar to the Wordy exercise?  Otherwise, it might be helpful to put a warning about needing the module.

kytrinyx:
Yeah, leaving them as optional/stretch-goal seems like a good compromise. What do you think @petemcfarlane?

petemcfarlane:
I think that sounds sensible. Do you suggesting making them stretch goals in the [x-common anagram.json](https://github.com/exercism/x-common/blob/deb225ea46f2c0c20bd240469eda6ef3366f7781/anagram.json#L108-L120) or just in xPHP track? Would you like to create a PR?

kytrinyx:
I think it would be worth making them a separate section in the JSON, and then each language track can decide whether or not it should be a stretch goal or not.

brettsantore:
Is there some documentation on how to handle stretch goals?

petemcfarlane:
I'm not sure, that's why I would like to ask @kytrinyx for advice 🙂

kytrinyx:
I don't think we have anything specific at the moment. We could put a comment above those tests. We could also add a `HINTS.md` in the exercise directory explaining why it's a stretch goal. The `HINTS.md` gets incorporated into the README when the exercise is fetched.

lafent:
Can I do anything to help here? I feel like I'm sort of pulling everyone's
focus and time here, the least I can do is pitch in. I'm just not sure
what.

petemcfarlane:
Hi @lafent, if you would like to help you could create a pull request with the changes @kytrinyx suggested, namely commenting out the tests that would rely on the `mb_string` functions/lib, and creating a `HINTS.md` file which explains 
> The tests that are commented out are a stretch goal. This is because they will probably depend on the mb_string extension to be implemented, which isn't installed with every version of PHP. 

or something like that?

PR #285, opened 2019-08-05
Anagram: Skip tests which require `mb_string` functions

arueckauer:
This PR provides changes discussed in #81.


petemcfarlane:
perfect, thanks

Issue #283, opened 2019-08-04
Namespaces and PSR-4 compliance


arueckauer:
One of the points I am missing in the exercises of this track is the concept of namespaces (PSR-4 compatibility). Therefore, I would like to discuss and propose the following change.

### Suggestion

Each exercise separates actual business logic from tests by having separate `src` and `test` directories. The `README.md` stays in the exercise's root folder. The example solution (and coming stub file) are going into `src` and the test file into `test`. 

> The terminating class name corresponds to a file name ending in .php. The file name MUST match the case of the terminating class name.

Since there is no definition for declaration of functions pertaining to the filename, two possible solutions are crossing my mind. Either name the file `functions.php` or `exercise-name.php` - both residing within the exercise's namespace. Here I am interested in suggestions, since I haven't used a file collecting various functions in years.

### Example

Taking Bob as an example, this would result in the following structure:

- `/src` → namespace `Exercism\Bob`
  - `Bob.php` (stub file)
  - `example.php` (example solution)
- `/test` → namespace `ExercismTest\Bob`
  - `BobTest.php`
- `README.md`

### Looking for feedback

My intention is to provide an even greater learning experience for the student within a modern (meaning current) environment.

It would be great to get some feedback on the suggestion over all and possible side effects I might not have fully considered yet (like implications on test automation etc.). Thank you!

dstockto:
I really like the introduction and consistent use of namespaces for the PHP track. As a test, I went ahead and modified my old bob code to follow the example above. I added a setupBeforeClass that performed a "require_once" on "../src/Bob.php". I also had to update my signatures of setup and setupBeforeClass to include the void return declaration, but after that it worked.

I also tried this on the Accumulator exercise (function based). In addition to adding the directories and adding namespaces, I had to also add the use statement:

`use function Exercism\Accumulate\accumulate;`

Similarly to Bob, I also added a `setupBeforeClass` method to require_once the file from ../src/Accumulate.php.  Once those were in place, I had all but one test passing. That test was the one that uses a static function to feed into accumulate function. I had to update the code to include the namespace in the string:

```
    public function testAccumulateUsingStaticMethod()
    {
        $this->assertEquals([5, 6], accumulate(['Hello', 'World!'], '\ExercismTest\Accumulate\Str::len'));
    }
```

[...]

PR #287, opened 2019-08-10
Update Accumulate to use namespaces

dstockto:
Split test classes to their own files

@arueckauer - Thoughts on these changes?

### DIFF
--- a/exercises/accumulate/example.php
+++ b/exercises/accumulate/example.php
@@ -1,11 +1,13 @@
 <?php
 
+namespace Exercism\Accumulate;
+
 /**
  * Given the input array and operation returns an array
  * containing the result of applying that operation
  * to each element of the input array.
  *
- * @param array    $input
+ * @param array $input
  * @param callable $accumulator
  *
  * @return array
--- a/exercises/accumulate/src/accumulate.php
+++ b/exercises/accumulate/src/accumulate.php
@@ -1,5 +1,7 @@
 <?php
 
+namespace Exercism\Accumulate;
+
 function accumulate(array $input, callable $accumulator)
 {
     // YOUR CODE GOES HERE
--- a/exercises/accumulate/test/AccumulateTest.php
+++ b/exercises/accumulate/test/AccumulateTest.php
@@ -1,10 +1,24 @@
 <?php
 
-class AccumulateTest extends PHPUnit\Framework\TestCase
+namespace ExercismTest\Accumulate;
+
+use PHPUnit\Framework\TestCase;
+use function Exercism\Accumulate\accumulate;
+
+class AccumulateTest extends TestCase
 {
     public static function setUpBeforeClass() : void
     {
-        require_once 'accumulate.php';
+        $files = [
+            '../src/accumulate.php',
+            'Str.php',
+            'StrSplitter.php',
+            'Is.php',
+        ];
+
+        foreach ($files as $file) {
+            require_once __DIR__ . '/' . $file;
+        }
     }
 
     public function testAccumulateEmpty() : void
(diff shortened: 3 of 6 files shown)


exercises/accumulate/test/accumulate_test.php
```diff
@@ -1,13 +1,27 @@
...
-        require_once 'accumulate.php';
+        $files = [
+            '/../src/accumulate.php',
+            '/Str.php',
+            '/StrSplitter.php',
+            '/Is.php',
+        ];
+
+        foreach ($files as $file) {
+            require_once $file;
```
arueckauer: This raises a warning

> [..] failed to open stream: No such file or directory [..]

I have to prepend the directory constant, no matter from which directory I'm running PHPUnit. I have not tested it with `make`, which I don't think works in this state. `require_once __DIR__ . $file;`

exercises/accumulate/example.php
```diff
@@ -1,11 +1,13 @@
 <?php
```
arueckauer: DId you leave the example file in the execise's root on purpose? Personally I would place it in `src`, because there I would look for it and it makes adjusting `Makefile` a little easier.

exercises/accumulate/test/accumulate_test.php
```diff
@@ -1,13 +1,27 @@
 <?php
```
arueckauer: Although we have no Autoloading, I'd prefer `AccumulateTest.php` as filename. Any cons on this?

exercises/accumulate/example.php
arueckauer: On second thought:

Not sure, if the exercise root folder would still be the better place. I have to sleep over it...

Is there an exercise, where the example solution requires multiple files (more than one class)? That would require some thought on how we can solve this. In that case an example directory might be the better solution.

exercises/accumulate/test/accumulate_test.php
dstockto: Fixed

exercises/accumulate/example.php
dstockto: Isn't that more part of how exercism works regardless of the language?

exercises/accumulate/test/accumulate_test.php
dstockto: None. I agree with this and will update.

exercises/accumulate/example.php
arueckauer: Not sure, I am mainly on the PHP track for now. We leave it here for now, that's fine.

dstockto:
@petemcfarlane We've made a couple of PRs here updating to using src and test directories as well as naming the files in a way that matches PHP PSR-4 standards. However, I didn't know or notice there's a makefile that specifies the names of the example and test file. I'm not sure how or if we should proceed with these. The Accumulate (this one) and Acronym PRs have been updated this way, but they won't work with the existing makefile and I'm not sure of a great way forward short of updating everything all at once. Any thoughts or ideas here?

arueckauer:
If I may add some additional information for clarification:

The goal of this PR is to provide a POC for #283. Other exercises would need an update as well. Preferably in one PR.

@petemcfarlane The primary question we have, is, what your thoughts are on this and whether you can support this effort? Or have doubts/concerns?

This PR currently fails test (Travis-CI was somehow not triggered). The issue is that the changed directory structure is not compatible with the Makefile configuration. We have two thoughts:

1. Update Makefile (together with the changes for all exercises) to use the new directory structure.
2. Introduce an `example` directory, which gets copied to `src` to run automated testing. This would be required, if multiple example solution files are required by an exercise.

dstockto:
If there's a way to do all this work where it doesn't require it to be in on PR, that would be preferred.  One PR doing all this would be massive and impossible to accurately review.

arueckauer:
Once we agree on a structure, it should be pretty straight forward. But again, I have not completed all exercises yet and don't know for sure if there is an exercise that raises the complexity in changing to namespaces.

It might be best to develop against a separate branch and then merge back to master as soon as everything is fine.

dstockto:
Closing this. I think much of what this was trying to accomplish has been handled elsewhere and more recently.

PR #288, opened 2019-08-10
Update acronym to namespace and data provider tests

dstockto:
@arueckauer - Here's a combined namespace and updating to data provider tests.


exercises/acronym/example.php
```diff
@@ -1,4 +1,5 @@
 <?php
+namespace Exercism\Acronym;
```
arueckauer: I've seen both a blank line before namespace declaration and none. Personally I am fine with either one, would be great if we can decide on one rule. 😊

exercises/acronym/example.php
```diff
@@ -8,13 +9,5 @@
...
-    $matches = preg_match_all('/\p{Lu}+\p{Ll}*|\p{Ll}+/u', $phrase, $words);
-
-    if ($matches === false || $matches < 2) {
-        return '';
-    }
-
-    return array_reduce($words[0], function ($acronym, $word) {
-        return $acronym . mb_strtoupper(mb_substr($word, 0, 1));
-    });
+    // Your code goes here
```
arueckauer: Probably mixed up content of example and stub?

exercises/acronym/test/acronym_test.php
```diff
@@ -0,0 +1,48 @@
+<?php
```
arueckauer: Also here: I would prefer `AcronymTest.php` as filename.

exercises/acronym/example.php
dstockto: Yup :-/

exercises/acronym/example.php
dstockto: I think I'll go with a blank before. Looking at PSR-2 it seems it's not needed or not specified, but when I set up PHP Storm to do PSR-2 formatting, I get the blank. So I'd rather go with what the computer automatically formatted than worry about manually making things consistent.

exercises/acronym/test/acronym_test.php
dstockto: Agreed.

dstockto:
Updated with suggestions above

dstockto:
I think for your other comments, I'd prefer we do the namespace with one blank line after the php tag, then the namespace. Then another blank and then the use statements.

For the return types, I'd prefer no space after the paren and then one space after the colon before the return type. 

For the test name, I agree for sure on the class base exercises - it should be named ClassNameTest.php for a ClassName.php class file. 

For function based, I think I'd go lower case since this shows that classes and their file names should be capitalized.

So it could be like this:

functions in function.php
tests in  function_test.php

What do you think?

dstockto:
Actually, I want to walk back part of the last bit. The tests are always classes. So the functions could be in function.php and the tests would be in FunctionTest.php.

exercises/acronym/test/AcronymTest.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace ExercismTest\Acronym;
+
+use PHPUnit\Framework\TestCase;
+use function Exercism\Acronym\acronym;
```
arueckauer: [PSR-12](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-12-extended-coding-style-guide.md) got approved. By its defintion, there needs to be a blank line between imports of classes and functions.

arueckauer:
> I think for your other comments, I'd prefer we do the namespace with one blank line after the php tag, then the namespace. Then another blank and then the use statements.

> For the return types, I'd prefer no space after the paren and then one space after the colon before the return type.

I'm okay with both rules. Can we define these in `phpcs-php.xml`? I am not too familiar with the configuration, but could do some reading. Just want to make sure, we use same code style.

About your suggestion to file names: Yes, this is a good suggestion.

exercises/acronym/test/AcronymTest.php
dstockto: I don't think it was approved yet when I made this PR. :)

arueckauer: You're right. 😋

dstockto:
> I'm okay with both rules. Can we define these in `phpcs-php.xml`? I am not too familiar with the configuration, but could do some reading. Just want to make sure, we use same code style.

Yes, that should be possible.

exercises/acronym/src/acronym.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Exercism\Acronym;
+
+/**
+ * Abbreviates a phrase.
+ *
+ * @param string $phrase
+ * @return string
+ */
+function acronym($phrase)
```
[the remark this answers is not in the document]
dstockto: In addition to being a fan of data providers I'm also all for type declarations and return types.

exercises/acronym/test/AcronymTest.php
```diff
@@ -0,0 +1,47 @@
...
+    // Additional points for making the following tests pass
+    public function testOneWordIsNotAbbreviated() : void
+    {
+        $this->assertEmpty(acronym('Word'));
+    }
+
+    public function testUnicode() : void
+    {
+        $phrase = 'Специализированная процессорная часть';
+        $this->assertEquals('СПЧ', acronym($phrase));
```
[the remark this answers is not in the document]
dstockto: I'm... not sure. It looks like it would fit just fine. I can only guess that it was because it was separated in the file from the others by the empty test.

exercises/acronym/test/AcronymTest.php
```diff
@@ -0,0 +1,47 @@
...
+            'camelCase'        => ['HTML', 'HyperText Markup Language'],
+            'all-caps words'   => ['PHP', 'PHP: Hypertext Preprocessor'],
+            'hyphenated'       => ['CMOS', 'Complementary metal-oxide semiconductor'],
+        ];
+    }
+
+    // Additional points for making the following tests pass
+    public function testOneWordIsNotAbbreviated() : void
+    {
+        $this->assertEmpty(acronym('Word'));
```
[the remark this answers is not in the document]
dstockto: I think that's another good point. That would allow all of these to be tested through the data provider.

exercises/acronym/test/AcronymTest.php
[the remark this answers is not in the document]
dstockto: Actually, I think it was due to the comments in the test - "extra points" if you also complete the tests under the comment".

exercises/acronym/test/AcronymTest.php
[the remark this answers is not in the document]
dstockto: I've added this and the other to the data provider.

exercises/acronym/src/acronym.php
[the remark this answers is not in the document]
dstockto: Added

PR #289, opened 2019-08-11
Improve gigasecond tests

arueckauer:
@dstockto These are the changes discussed in #276 

- Remove no mutation test
- Add assertion instance of `DateTimeImmutable`

### DIFF
--- a/exercises/gigasecond/gigasecond_test.php
+++ b/exercises/gigasecond/gigasecond_test.php
@@ -2,18 +2,18 @@
 
 class GigasecondTest extends PHPUnit\Framework\TestCase
 {
-    public static function setUpBeforeClass(): void
+    public static function setUpBeforeClass() : void
     {
         require_once 'gigasecond.php';
     }
 
-    public function dateSetup($date): DateTimeImmutable
+    public function dateSetup($date) : DateTimeImmutable
     {
         $UTC = new DateTimeZone('UTC');
         return new DateTimeImmutable($date, $UTC);
     }
 
-    public function testForDateOnlySpecificationOfTime(): void
+    public function testForDateOnlySpecificationOfTime() : void
     {
         $date = $this->dateSetup('2011-04-25');
         $gs = from($date);
@@ -22,7 +22,7 @@ class GigasecondTest extends PHPUnit\Framework\TestCase
         $this->assertInstanceOf(DateTimeImmutable::class, $gs);
     }
 
-    public function testForDateOnlySpecificationOfTimeTwo(): void
+    public function testForDateOnlySpecificationOfTimeTwo() : void
     {
         $date = $this->dateSetup('1977-06-13');
         $gs = from($date);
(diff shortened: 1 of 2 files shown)


exercises/gigasecond/gigasecond_test.php
```diff
@@ -7,68 +7,66 @@ public static function setUpBeforeClass() : void
...
-    public function dateSetup($date) : \DateTimeImmutable
+    public function dateSetup($date) : DateTimeImmutable
     {
-        $UTC = new DateTimeZone("UTC");
+        $UTC = new DateTimeZone('UTC');
         return new DateTimeImmutable($date, $UTC);
     }
 
-    public function testDate1() : void
+    public function testForDateOnlySpecificationOfTime() : void
```
dstockto: Do you prefer the spaces before and after the colon? I can adjust my formatter for that. I've just always used no space before and space after.

arueckauer: Either way is fine. I removed it.

exercises/gigasecond/gigasecond_test.php
```diff
@@ -7,68 +7,66 @@ public static function setUpBeforeClass() : void
...
-
-        $this->assertNotEquals($date, $gs);
+        $this->assertSame('2046-10-03 01:46:39', $gs->format('Y-m-d H:i:s'));
+        $this->assertInstanceOf(DateTimeImmutable::class, $gs);
     }
 
     public function testYourself() : void
     {
-        // Replace the string "your_birthday" with your birthday's datestring
+        // Replace the string "your_birthday" with your birthday's date string and remove line 65.
```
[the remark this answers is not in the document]
arueckauer: Agreed. Removed it.

exercises/gigasecond/gigasecond_test.php
```diff
@@ -7,68 +7,54 @@ public static function setUpBeforeClass() : void
...
+    public function testFullTimeWithDayRollOver() : void
     {
-        $date = $this->dateSetup("2015-01-24 23:59:59");
+        $date = $this->dateSetup('2015-01-24 23:59:59');
         $gs = from($date);
 
-        $this->assertSame("2046-10-03 01:46:39", $gs->format("Y-m-d H:i:s"));
-    }
-
-    public function testNoMutation() : void
```
petemcfarlane: Instead of removing this test and asserting in every other test instance that `from` will return a `DateTimeImmutable`, could we just keep this test, somewhere near the start perhaps?

petemcfarlane: Another option is to use a PHPUnit data provider - that might tidy things up a bit, and we could just assert in one place that the type is a `DateTimeImmutable`?

dstockto: I'm all for data providers when we can.

arueckauer:
Thanks for the review. I am on vacation until mid-September and will have a look at it then.

exercises/gigasecond/gigasecond_test.php
arueckauer: Excellent input. Added data providers in 42c620e.

arueckauer:
@petemcfarlane Any more feedback on this change or is it ok to merge?

PR #298, opened 2019-11-29
Improve readme for gigasecond exercise

MatheusNaldi:
The hardest step of this exercise was to understand what I have to do.

### DIFF
--- a/exercises/gigasecond/README.md
+++ b/exercises/gigasecond/README.md
@@ -1,6 +1,6 @@
 # Gigasecond
 
-Calculate the moment when someone has lived for 10^9 seconds.
+You must add a gigasecond to the informed date.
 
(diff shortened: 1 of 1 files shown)


MatheusNaldi:
> Fair enough 
> The canonical problem specification description for the exercise isn't much clearer either - you might want to submit a PR to there as well? https://github.com/exercism/problem-specifications/blob/master/exercises/gigasecond/description.md

Where in the exercism.io I can access this problem specifications? I've been searching but I couldn't find it. It would help me.

petemcfarlane:
Hi @MatheusNaldi I'm not sure where in the app they are displayed or if they're just used for track maintainers to use as examples, perhaps @kytrinyx can help clarify?

MatheusNaldi:
I found it, it is in the exercises list:
image

PR #300, opened 2020-02-04
Clarify how to run the tests

N-Parsons:
It wasn't initially obvious to me how to run the tests from the information given - I was attempting to run it with `phpunit hello-world.php`. The error message that this generates isn't very useful, so this PR helps new users by giving a specific example of how to run the tests.


neenjaw (approved):

neenjaw:
Long delay, but thanks for this suggestion! 💙

PR #306, opened 2020-04-01
Update dependencies

petemcfarlane:
- PHPUnit 9 currently doesn't work with our autoloading/class name strategy - this will need more consideration (so specify PHPUnit 8)
- The codesniffer version we were using is several years out of date and has errors with PHP 7
- Also include PHP 7.4 support, and drop PHP 7.2 support (replaces #304 )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 ---
 language: php
 php:
-- 7.2
 - 7.3
+- 7.4
 script:
 - make install
 - make style-check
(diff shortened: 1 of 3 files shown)


petemcfarlane:
I'm not sure anyones actively reviewing PRs here, but this does get the CI build passing again, and will allow other PRs to be unblocked and merged, so I'll merge this now, then if anyone has any issues please create an issue or comment on this PR and we can review/fix then 🙂

PR #334, opened 2021-02-04
WIP - [v3] Upgrading PHPCS ruleset to PSR-12

neenjaw:
This is a work in progress to upgrade the code standard to PSR-12


neenjaw:
This will still have failing checks for PHPUnit 9, as that is still not fixed on the track.  All of checks should be passing.

ErikSchierboom (approved):
LGTM

PR #344, opened 2021-02-28
Empty concept exercises

djjohnsongeek:
Added empty concept exercise files based on [this list.](https://exercism-team.slack.com/archives/CSPV0K5GT/p1613006298000500) Let me know if anything needs to be fixed!


neenjaw:
First off, this is awesome!

I think this provides a great scaffold for us to plug things in to the repo and to get it done piece by piece.

I changed the base to a branch just because it fails our CI. Then as we finish sections we can cherry pick commits to a new branch, then merge those into master.

So I would propose that we merge this then:
- start filling in concepts
  - these can be pretty bare bones, but just to have something in them
- start by making exercises
  - try to keep them small and focal to one learning point at a time
  - the Lasagna (or similar) exercise would be a good choice to start with, be creative or fork them from other tracks, both strategies are good

For concept exercises, as we complete them, we'll need to change their containing folder to a name that reflects the exercise and not the concept.  eg. If it was `exercises/concept/functions` it would migrate to `exercises/concept/lasagna` or some-such.

Thoughts?

djjohnsongeek:
Sounds like a good plan to me!

PR #349, opened 2021-03-09
Variable concepts

neenjaw:


concepts/variable-scope/about.md
```diff
@@ -0,0 +1,42 @@
+# About
```
ErikSchierboom, line 1: You've explained these concepts very well!

ErikSchierboom (approved):
Excellent work

PR #353, opened 2021-03-21
all-your-base: Flip argument order in test assertions to match PHPUnit expectations

mk-mxp:
I just stumbled over the tests telling me unexpected things about what was wrong. It turned out to be caused by the order of `$actual,$expected` in the tests, instead of PHPUnit's `$expected, $actual`. This pull request flips that to PHPUnit's order.


neenjaw (approved):
Awesome catch! Thanks for submitting a PR to fix this quality of life change!

You won't see it go live on the website until after v3 launches, but I can guarantee your contribution will make a difference to others in the future!

Issue #347, opened 2021-03-07
Exercise stubs


neenjaw:
Many exercises don't have stubs.  For example Gigasecond is an empty file, but it should have a stub defined like:

```php
// gigasecond.php
<?php

function from(DateTimeImmutable $date) {
    throw new Exception('Not implemented');
}
```

This will lead to a better experience using the website code editor and also help to define the sandbox of solutions for students.

codedge:
Is that something where I can through the exercises and check? [Here](https://github.com/exercism/php/pull/351#discussion_r598207310) we used a `BadFunctionCallException`

```php
<?php

function xyz(int $number): string {
  throw new \BadFunctionCallException("Implement the xyz function");
}
```

So can we agree on using this exception throughout all the stubs?

neenjaw:
In exercises that are just functions, yes, I think that exception is the proper one to use (PHP doesn't have a native `NotImplementedException` and I would like to keep exercises simple).  For exercises which require a class (I'm not sure which ones off-hand) it should probably be something like:

```php
<?php

class ExampleClass
{
    public function doSomething(): int
    {
        throw new \BadMethodCallException("Implement the doSomething method");
    }
}
```

neenjaw:
Fulfilled by #357

PR #357, opened 2021-03-24
Provide stubs

codedge:
Create exercise stubs, see #347 

This is just the first set, I need to go through a couple of others. It seems to be the minority that have stubs.

### DIFF
--- a/exercises/practice/accumulate/Accumulate.php
+++ b/exercises/practice/accumulate/Accumulate.php
@@ -1,6 +1,6 @@
 <?php
 
-function accumulate(array $input, callable $accumulator)
+function accumulate(array $input, callable $accumulator): array
 {
-    // YOUR CODE GOES HERE
+    \BadFunctionCallException("Implement the accumulate function");
 }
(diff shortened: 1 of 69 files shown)


codedge:
I ran through all the exercises adding stubs. Should be good now.

neenjaw:
@codedge - Thanks for opening this PR! I'll review this tonight. From a brief skim, this looks great!

exercises/practice/accumulate/Accumulate.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-function accumulate(array $input, callable $accumulator)
+function accumulate(array $input, callable $accumulator): array
 {
-    // YOUR CODE GOES HERE
+    \BadFunctionCallException("Implement the accumulate function");
```
neenjaw, line 5: Shouldn't these throw the error?
```suggestion
    throw new \BadFunctionCallException("Implement the accumulate function");
```

neenjaw (changes requested):

exercises/practice/accumulate/Accumulate.php
codedge: Holy... your are completely right. Fixed for all the tests.

neenjaw (approved):
Very nice! Awesome work @codedge!

PR #359, opened 2021-03-26
Add blurb

zembrowski:
Ordinal number meta config misses the "blurb" key and therefore the Configlet CI is failing.


neenjaw (approved):
Thanks for adding this in! 💙

PR #363, opened 2021-04-16
Change return statement

zembrowski:


exercises/practice/accumulate/AccumulateTest.php
```diff
@@ -111,6 +111,6 @@ class Is
 {
     public function truthy($value): bool
     {
-        return $value ? true : false;
+        return (bool) $value;
```
neenjaw, line 114: what do you think of using `boolval` rather than casting it to a boolean?

zembrowski: According to the following three sources, it is better to use casting instead of *val() functions:

1. [PHP: floatval](https://www.php.net/manual/en/function.floatval.php#123303) comment

2. PhpStorm plugin Php Inspections ​(EA Extended)
ea

3. [What's the difference between $bar = boolval($foo) and $bar = (bool) $foo in php?](https://stackoverflow.com/a/34653513/7667704) 

In this particular case there is no real impact difference.

I prefer casting.

neenjaw: Here we are converting values to a boolean, so I don't think the comparison to convert to floating point values is easily comparable. I think the opinion about cognitive load isn't quite founded, since many other languages use a functional construct to convert values from one type to another (e.g. JavaScript).

PHP 4 and 5 are hugely out of date and unsupported, so we should not be modifying any code to teach based on those versions.  I have benchmarked these two methods in php 7.4.15 and 8.0.2 and found negligible difference between each method (php 8's performace improves vastly over 7.4):

> My [benchmarking script here in this gist](https://gist.github.com/neenjaw/e5390a187c2d742cb458ae6a3a5d9960)

| Key | Value |
|--- |--- |
|PHP Version|7.4.15|
|casting true|0.045 sec.|
|casting false|0.040 sec.|
|boolval true|0.038 sec.|
|boolval false|0.039 sec.|
|Total|0.162 sec.|

| Key | Value |
|--- |--- |
|PHP Version|8.0.2|
|casting true|0.021 sec.|
|casting false|0.020 sec.|
|boolval true|0.018 sec.|
|boolval false|0.019 sec.|
|Total|0.077 sec.|


Lastly, I don't prefer casting, but I don't think any preferences really matter.  We should teach idiomatic and performant php with good practices. Which means benchmarking our opinions and seeing if they hold true. In this case we need to make sure that we are pushing students towards php 8 and `boolval` even if it is negligible.

Please run the script locally for you to make sure that the benchmarking is correct.  If there is a significant variance it would be interesting to know.

neenjaw:
```suggestion
        return bollval($value);
```

zembrowski: Valid arguments. Hard to argue with them.

I run the benchmark locally (7.4/8.0 with opcache disabled or enabled) and must admit the differences are marginal for me (cf. example below).

Key |  Value
-- | --
PHP Version | 8.0.3
casting true | 0.180 sec.
casting false | 0.180 sec.
boolval true | 0.181 sec.
boolval false | 0.180 sec.

Currently almost none casts are used in examples/tests and multiple *val() functions are being used, therefore I changed this PR to the suggested boolval().

neenjaw (approved):

PR #383, opened 2021-06-27
Add strict types for all php files

### LINKED ISSUE #352: Use strict types
Hey @petemcfarlane, @djjohnsongeek, @arueckauer,

as discussed with @neenjaw in a [new exercise](https://github.com/exercism/php/pull/351#discussion_r598208916) an enhancement to the php tasks would be to enforce strict types by adding `declare(strict_types=1);` to every php file.

This should be already in the skeleton files offered by the exercise.io platform, so that a possible file could look like this:

```php
<?php

declare(strict_types=1);

function xyz() {

}
```

What do you think?

codedge:
Resolve #352 

This PR adds `declare(strict_types=1)` to all php files and an explanation what it means.

The sentences of the explanation are sometimes a bit long to match the 120 chars limit that is enforces with the PSR-2 standard. I therefore made some line breaks to meet this requirement.

### Rule to enfore strict types

An additional phpcs rule enforces a proper declaration of strict types for each file. The rule [`SlevomatCodingStandard.TypeHints.DeclareStrictTypes`](https://github.com/slevomat/coding-standard#slevomatcodingstandardtypehintsdeclarestricttypes-) from [`slevomat/coding-standards`](https://github.com/slevomat/coding-standard) takes care of this.

### Rule to enforce explanation of strict type

I've added a custom phpcs rule, that takes care of the explanation. I think this will make it easier to apply coding standards in the future.

### Using composer for lint check

To properly include the custom rules, [composer](https://getcomposer.org/) is used to run the check in Github actions. 

### Specifying php version

To properly specify the version used to CI action, the key `php-version` needs to be used, instead of `version`. See [`shivammathur/setup-php`](https://github.com/shivammathur/setup-php)

### DIFF
--- a/exercises/practice/accumulate/.meta/example.php
+++ b/exercises/practice/accumulate/.meta/example.php
@@ -3,13 +3,14 @@
 /*
  * By adding type hints and enabling strict type checking, code can become easier to read,
  * self-documenting and reduce the number of potential bugs.
- * By default, type declarations are non-strict,
- * which means they will attempt to change the original type to match the type specified by the type-declaration.
+ * By default, type declarations are non-strict, which means they will attempt to
+ * change the original type to match the type specified by the type-declaration.
  * In other words, if you pass a string to a function requiring a float,
  * it will attempt to convert the string value to a float.
  * To enable strict mode, a single declare directive must be placed at the top of the file.
  * This means that the strictness of typing is configured on a per-file basis.
- * This directive not only affects the type declarations of parameters, but also a function's return type.
+ * This directive not only affects the type declarations of parameters, but also a function's
+ * return type.
  *
  * For more info review the Concept on strict type checking in the PHP track <link>.
  * To disable strict typing, comment out the directive below.
--- a/exercises/practice/accumulate/.meta/example.php
+++ b/exercises/practice/accumulate/.meta/example.php
@@ -1,18 +1,24 @@
 <?php
 
 /*
- * By adding type hints and enabling strict type checking, code can become easier to read,
- * self-documenting and reduce the number of potential bugs.
- * By default, type declarations are non-strict, which means they will attempt to
- * change the original type to match the type specified by the type-declaration.
+ * By adding type hints and enabling strict type checking, code can become
+ * easier to read, self-documenting and reduce the number of potential bugs.
+ * By default, type declarations are non-strict, which means they will attempt
+ * to change the original type to match the type specified by the
+ * type-declaration.
+ *
  * In other words, if you pass a string to a function requiring a float,
  * it will attempt to convert the string value to a float.
- * To enable strict mode, a single declare directive must be placed at the top of the file.
- * This means that the strictness of typing is configured on a per-file basis.
- * This directive not only affects the type declarations of parameters, but also a function's
- * return type.
  *
- * For more info review the Concept on strict type checking in the PHP track <link>.
+ * To enable strict mode, a single declare directive must be placed at the top
+ * of the file.
+ * This means that the strictness of typing is configured on a per-file basis.
+ * This directive not only affects the type declarations of parameters, but also
+ * a function's return type.
+ *
+ * For more info review the Concept on strict type checking in the PHP track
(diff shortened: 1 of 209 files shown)


neenjaw:
If this is WIP can we make it a draft PR? Looks good so far, just needs the explanatory message

codedge:
Yes, please make it a draft. I know there are some things open and I'll ping you when I think it is ready.

codedge:
@neenjaw I feel like the PR is ready to be checked. All changes should be explained in my comment above, I hope 😉

neenjaw:
Cool, I'll check it out later today

neenjaw (approved):
One small comment about the formatting of the paragraph.  Looks great otherwise.  Tell me what you think and then I can merge?

codedge:
Yes, good idea to go for the 80 chars. I am gonna change it and then you can merge.

codedge:
@neenjaw I tried to justify the text a bit more but it is hard to not "destroy" the sentence with unneeded line breaks. Do you think it is ok so?

exercises/practice/accumulate/.meta/example.php
```diff
@@ -1,5 +1,23 @@
...
+ * In other words, if you pass a string to a function requiring a float,
+ * it will attempt to convert the string value to a float.
+ * To enable strict mode, a single declare directive must be placed at the top of the file.
+ * This means that the strictness of typing is configured on a per-file basis.
+ * This directive not only affects the type declarations of parameters, but also a function's
+ * return type.
+ *
+ * For more info review the Concept on strict type checking in the PHP track <link>.
+ * To disable strict typing, comment out the directive below.
+ */
```
SleeplessByte, line 17:
```suggestion
/*
 * By adding type hints and enabling strict type checking, code can become 
 * easier to read, self-documenting and reduce the number of potential bugs.
 * By default, type declarations are non-strict, which means they will attempt
 * to change the original type to match the type specified by the 
 * type-declaration. In other words, if you pass a string to a function 
 * requiring a float, it will attempt to convert the string value to a float.
 * 
 * To enable strict mode, a single declare directive must be placed at the top
 * of the file. This means that the strictness of typing is configured on a 
 * per-file basis. This directive not only affects the type declarations of
 * parameters, but also a function's return type.
 *
 * For more info review the Concept on strict type checking in the PHP track 
 * <link>.
 * 
 * To disable strict typing, comment out the directive below.
 */
```

This would be a strict 80 width adaptation.

SleeplessByte, line 17:
```suggestion
/*
 * By adding type hints and enabling strict type checking, code can become 
 * easier to read, self-documenting and reduce the number of potential bugs.
 * By default, type declarations are non-strict, which means they will attempt 
 * to change the original type to match the type specified by the 
 * type-declaration.
 * 
 * In other words, if you pass a string to a function requiring a float,
 * it will attempt to convert the string value to a float.
 * 
 * To enable strict mode, a single declare directive must be placed at the top 
 * of the file.
 * This means that the strictness of typing is configured on a per-file basis.
 * This directive not only affects the type declarations of parameters, but also 
 * a function's return type.
 *
 * For more info review the Concept on strict type checking in the PHP track 
 * <link>.
 * 
 * To disable strict typing, comment out the directive below.
 */
```

Here is a strict 80 width adaptation based on your line breaks.

SleeplessByte: I've also tried 120 chars width, but that looks really bad.

neenjaw:
Thanks @SleeplessByte, I think one of those formats would be great.  @codedge, which formatting do you think works best for this use?

exercises/practice/accumulate/.meta/example.php
SleeplessByte: (I recommend copying and pasting it in your IDE if you want to preview how it looks)

codedge:
I change the text to match the second suggestion. Looks better now?

neenjaw:
Looks great, let's merge it

PR #393, opened 2021-07-25
Add Poker exercise to PHP Track.

MichaelBunker:
### A few important notes.

- The README is intentionally small because it matches the READMEs for this exercise from other tracks.
- The example solution is fairly long (200+ lines). I'm open to recommendations on trimming it down if possible since it's fairly large, However, I think it strikes the right balance between readability and succinctness given the challenge requirements.
- There are multiple classes in the example file. I personally tend to avoid having more than one class in a single file, but I don't know what this would do in the Exercism ecosystem if we had multiple files for an example. Having classes instead of stdClass objects helps readability AND I think including them in the single example file will help readability and simplicity for students/mentors if they end up referencing the example.

#### Instructions

Pick the best hand(s) from a list of poker hands.

See [wikipedia](https://en.wikipedia.org/wiki/List_of_poker_hands) for an overview of poker hands.

### DIFF
--- a/config.json
+++ b/config.json
@@ -802,6 +802,19 @@
         ]
       },
       {
+        "slug": "poker",
+        "name": "Poker",
+        "uuid": "e8669c8e-bfdd-4207-baf6-23c2e4b12d13",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 5,
+        "topics": [
+          "games",
+          "parsing",
+          "sorting"
+        ]
+      },
+      {
         "slug": "palindrome-products",
         "name": "Palindrome Products",
         "uuid": "3201ddeb-9046-4801-a31a-28c292686e1e",
--- a/exercises/practice/poker/.docs/instructions.md
+++ b/exercises/practice/poker/.docs/instructions.md
@@ -1,0 +1,5 @@
+# Instructions
+
+Pick the best hand(s) from a list of poker hands.
+
+See [wikipedia](https://en.wikipedia.org/wiki/List_of_poker_hands) for an overview of poker hands.
--- a/exercises/practice/poker/.meta/config.json
+++ b/exercises/practice/poker/.meta/config.json
@@ -1,0 +1,10 @@
+{
+  "blurb": "Pick the best hand(s) from a list of poker hands.",
+  "authors": ["MichaelBunker"],
+  "contributors": [],
+  "files": {
+    "solution": ["Poker.php"],
+    "test": ["PokerTest.php"],
+    "example": [".meta/example.php"]
+  }
+}
--- a/exercises/practice/poker/.meta/example.php
+++ b/exercises/practice/poker/.meta/example.php
@@ -1,0 +1,218 @@
+<?php
+
+/*
+ * By adding type hints and enabling strict type checking, code can become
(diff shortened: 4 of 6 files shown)


MichaelBunker:
It appears the Configlet CI check is failing

>  The value of `key_features.icon` is the empty string, but it must be one of the allowed values:
/home/runner/work/php/php/config.json

Running unit tests and linters locally (and the matching CI checks for this PR) are passing. @neenjaw Is this configlet CI check something I need to resolve?

neenjaw:
> It appears the Configlet CI check is failing
> 
> >  The value of `key_features.icon` is the empty string, but it must be one of the allowed values:
> /home/runner/work/php/php/config.json
> 
> Running unit tests and linters locally (and the matching CI checks for this PR) are passing. @neenjaw Is this configlet CI check something I need to resolve?

No, this is a new site-wide configlet check that was merged today. If it is the only thing failing, can resolve in it a separate PR

neenjaw:
@MichaelBunker this exercise needs an entry in the config.json with a unique v4 uuid.

@petemcfarlane, can you code review the example? Aside from the lacking config entry, the files seem good from a v3/test runner perspective

MichaelBunker:
> 
> 
> @MichaelBunker this exercise needs an entry in the config.json with a unique v4 uuid.


Other tracks don't seem to have a consensus on the difficulty or topics for this exercise so I used my best judgement. If there are some standards for how to define the config (how to determine difficulty, topics, etc) let me know.

Also, I recently added the Darts exercise and that didn't include an entry in the config. Is that discrepancy between the two exercises intentional?

```json
  {
    "slug": "poker",
    "name": "Poker",
    "uuid": "e8669c8e-bfdd-4207-baf6-23c2e4b12d13",
    "practices": [],
    "prerequisites": [],
    "difficulty": 5,
    "topics": [
      "games",
      "parsing",
      "sorting"
    ]
  }

petemcfarlane (approved):
@neenjaw I was under the impression the example files aren't necessarily an exemplar or optimal solution to the problem, rather just there to make sure that a solution exists and that the test cases work?
Having said that, after briefly reviewing, the example looks perfectly fine, it's pretty readable, uses sensible naming. There's possibly a few different stylist thing I might have done differently, but that's the joys of submitting solutions and getting feedback on Exercism ;) 

I'm OK with having multiple classes defined in the single file here, again it's only an example, the user doesn't have to do this. So far as I'm aware we don't have any other multi-file examples for other exercises? I'm not sure if this would work or be possible without changing the configlet?

Yes, having the `Card`, `Hand`, `Score` value objects are a bit more readable than just using `stdObject`, though we're not actually adding any behaviour/methods so they could easily be refactored to `stdObject`s - it's more of an opinion.

neenjaw:
@petemcfarlane, yep exactly. Example solutions for practice exercises are just about proving that it is solvable with the language. I think should still be a reasonable solution, but I am satisfied if you are satisfied. 

Thanks for reviewing the code, I am on vacation away from a computer so is a bit hard to review code well from a distance. 

If the only failing check is a configlet check about the icons, it should be good to merge if you see fit

petemcfarlane:
Ok, merging then!

PR #296, opened 2019-10-30
[WIP] Adds the Simple Cipher exercise

camilopayan:

PR #395, opened 2021-07-27
Add the Simple Cipher exercise.

MichaelBunker:
This PR contains the original commit from https://github.com/exercism/php/pull/296 as well as the work required to bring it in line with the current exercise requirements.

### Simple Cipher
It is a type of substitution cipher in which each letter in the plaintext is replaced by a letter some fixed number of positions down the alphabet. For example, with a left shift of 3, D would be replaced by A, E would become B, and so on.


petemcfarlane (approved):
Great, thanks everyone involved!

PR #295, opened 2019-10-29
Adds the High Scores exercise.

camilopayan:

PR #398, opened 2021-07-29
Add the High Scores exercise.

MichaelBunker:
This PR contains the original commit from https://github.com/exercism/php/pull/295 as well as the work required to bring it in line with the current exercise requirements.

### High Scores

Your task is to build a high-score component of the classic Frogger game, one of the highest selling and addictive games of all time, and a classic of the arcade era. Your task is to write methods that return the highest score from the list, the last added score and the three highest scores.


neenjaw (approved):
Nice work! Thanks for continuing @camilopayan's PR and bringing it up to speed, @MichaelBunker!

PR #399, opened 2021-07-29
Add the linked-list exercise.

MichaelBunker:
Added the linked-list exercise.

### Instructions

Implement a doubly linked list. You will write an implementation of a doubly linked list. Implement a Node to hold a value and pointers to the next and previous nodes. Then implement a List which holds references to the first and last node and offers an array-like interface for adding and removing items:

* `push` (*insert value at back*);
* `pop` (*remove value at back*);
* `shift` (*remove value at front*).
* `unshift` (*insert value at front*);

To keep your implementation simple, the tests will not cover error conditions.

### DIFF
--- a/config.json
+++ b/config.json
@@ -848,6 +848,18 @@
         ]
       },
       {
+        "slug": "linked-list",
+        "name": "Linked List",
+        "uuid": "e5f10431-8312-4d6c-abfa-332cd7beea42",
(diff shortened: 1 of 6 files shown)


neenjaw (approved):
Nice work, clean textbook implementation.  Appreciate your work, @MichaelBunker!

PR #401, opened 2021-07-30
Add the matrix exercise.

MichaelBunker:
Given a string representing a matrix of numbers, return the rows and columns of that matrix.


neenjaw (approved):
Nice work, thanks for implementing this problem, @MichaelBunker

PR #406, opened 2021-08-01
Add dnd-character exercise.

MichaelBunker:


neenjaw (commented):
This example solution prompts a question that I think we should probably address about how we should design the interface to this problem.

The current interface directly access class properties.  Not sure about what is "normal" in the wider php community, but my professional experience with PHP classes is that these generally would be private properties, accessed through getters.  Thoughts about this @petemcfarlane, @MichaelBunker?

MichaelBunker:
My experience before typed properties (pre 7.4) was private properties with getters and setters to enforce types.

After typed properties were introduced, I haven't felt the same need for getters and setters unless they are doing validation/manipulation or have some specific authorization for access of the input. Since none of them apply here, I didn't think it's necessary. My thought writing it was 'if this was a real world program, the stats of a character would be public (they would have to be for stat checks and rolls and what not in D&D), so make the visibility public. 

Let me know if it should be changed.

neenjaw:
🤷🏻 I don't have much of an opinion from that perspective for this limited use-case.  I think if the design of this were to optionally include some sort of computed property I would lean stronger towards the getter pattern.  That said, with that reasoning, I am good with it, but I'd like to know what @petemcfarlane thinks when he has a moment

neenjaw (approved):

petemcfarlane (approved):
I largely agree with @MichaelBunker, that it makes sense the properties are `public` for access/readability. Unfortunately PHP hasn't yet implemented `readonly` or `immutable` attributes for class properties, which means that the properties could be changed after the class here is constructed. At least with property typing it must be set to an `int` in this case. 

I think, in this exercise, the test is clear and simple and it saves the bloat of adding getter methods which add little value (as they're not doing any manipulation/transformation/validation).

neenjaw:
Sounds good to me. :)

PR #407, opened 2021-08-01
Add the reverse-string exercise.

MichaelBunker:


neenjaw (approved):
I think because the problem specification for this doesn't specify not to use built-in functions, this is acceptable.  Do you think this exercise covers a use-case missing from the track as is?

I ask because on other thracks I am involved in, we have had some discussions to specifically not include some problems from the `exercism/problem-specifications` repo because they are relatively uninteresting or because they are poorly designed (imho, `rest-api` is an example of poor design).

I think this problem is just relatively uninteresting because there is a built-in and serves mostly as a 'gotcha' because someone may or may not know about `strrev`.

Interested in your thoughts on this, @petemcfarlane, @MichaelBunker.  Happy to merge unless either of you otherwise say so.

MichaelBunker:
I've actually been giving this very issue a lot of thought lately because I've been deep in the problem specification repository AND also the C#, Go and Ruby repositories for inspiration and some issues related to this have popped up.

I agree with you on the idea that some lessons are just inherently less interesting because of the specific languages abilities, but I'm not as convinced that is a good reason not to include them. Knowing when to reach to the standard library for existing functionality is an important skill for professional developers, and the lessons that are 'trivial' will be boring for mentors, but could provide some insight for students who don't realize how extensive the feature set is for the standard library in PHP. These trivial exercises will at least keep that idea in their minds of 'can PHP handle this with the standard library' and that alone will make them think about, and engage with, the language - and that can only be viewed as a positive thing.

_**_If a student does this exercise and does it manually (and not use strrev()), a mentor might say 'good solution, but the standard library also supports easy string manipulation by calling strrev().' When the student encounters more exercises with strings, they will be encouraged to think about how the standard library can help them with string manipulation._**_ 

As an example - here is a function from the Matrix exercise. explode() and array_map() are powerful tools when used appropriately, and students should be encouraged to be thinking about how the language can help them complete their tasks.
```php
    private function parseRows(string $matrix): void
    {
        foreach (explode('\n', $matrix) as $stringRow) {
            $this->rows[] = array_map(fn(string $value) => (int) $value, explode(' ', $stringRow));
        }
    }
```

I think a very valuable lesson for programming languages is how powerful the standard library is for that language. I think this is more evident in other languages like go that specifically point to their standard libraries as great pieces of work that are expressive, easy to use, etc. In fact PHP is usually brought down a peg because of inconsistencies with it's standard library (for example, inconsistent needle and haystack parameter order). There are languages, specifically lower level languages "closer to the metal" where you don't get as many built in easy features from the standard library. 

I think that it should be embraced when languages DO support those things. Is it simple and boring coding strrev()? Yes. But I think it would be beneficial for students to be thinking about "I'm trying to do x Task in this language, Does the language support a simple straightforward solution to accomplish this task? If not, can some part of it at least be accomplished with the standard library?" And if the answer is a function in the standard library, why not embrace that?

The language tracks should be embracing the things each individual language is built for. Coroutines for example, are not a natively supported common feature in PHP. So we wouldn't expect exercises to have a simple solution for them in the exercises. Kotlin however, probably would have exercises that are solved "simply" by using supported coroutines.

The following is very opinionated and might not match what Exercism is striving for but my ultimate take is this:

- **_Students take courses to learn a specific language, and part of learning a specific language is learning the standard tools and features of that language. So we should embrace exercises that are solved simply with the standard library because they encourage students to think about the tools for that specific language (as long as there aren't an overwhelming majority of them)._**

I hope this was all clear and concise.

With all that being said, if the PHP maintainers feel it shouldn't be included, feel free to close this PR.

neenjaw:
Yea, I think that's a well-formed argument which I can align with. Thanks for your thoughts @MichaelBunker!

PR #412, opened 2021-09-11
Use single quote and at the end of the array

patrykwozinski:
Well, yes - it's a good practice to use a single quote character when working with simple strings. Also, another good habit is to use `,` for every element of an array (multi-line).

### DIFF
--- a/exercises/practice/resistor-color/ResistorColorTest.php
+++ b/exercises/practice/resistor-color/ResistorColorTest.php
@@ -33,32 +33,32 @@ class ResistorColorTest extends PHPUnit\Framework\TestCase
 
     public function testBlackColorCode(): void
     {
-        $this->assertEquals(colorCode("black"), 0);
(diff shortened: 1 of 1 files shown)


neenjaw:
Hi @patrykwozinski, we don't currently have a style guide to measure improvements like this against.  We have started talking about specific formatting/linting rules with regard to style (eg. #418 and #421) but I think that until we come to a consensus as a group of maintainers this is more of a preference.

If you could add your thoughts re: style to #418 I would appreciate that

NhanAZ:
> @patrykwozinski Can you elaborate, why it is good practice to use single quotes? What is the benefit here?

I think `'string'` is better but only when absolutely necessary
In this case, it is not necessary
`"string"` is better.

neenjaw:
Given that this is a stylistic choice and we have not settled on a unified coding style for this repo/track, I am closing.  A coding style/standard should be aligned upon in an issue before applying stylistic changes to individual exercises.

PR #415, opened 2021-09-27
Update tests and example to use actual newlines

dstockto:
Single-quoted \n is just that character, so the actual
output was just a single long line. This will make the output
look more like the instructions say it should and reduce confusion
when implementing.

### DIFF
--- a/exercises/practice/tournament/.meta/config.json
+++ b/exercises/practice/tournament/.meta/config.json
@@ -1,7 +1,7 @@
 {
   "blurb": "Tally the results of a small football competition.",
   "authors": ["MichaelBunker"],
-  "contributors": [],
+  "contributors": ["dstockto"],
   "files": {
     "solution": ["Tournament.php"],
     "test": ["TournamentTest.php"],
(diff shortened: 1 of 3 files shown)


petemcfarlane (approved):

petemcfarlane:
> [...] I disagree with the changes from `+=` to `++`.

@neenjaw aren't they practically the same here? I suppose it would be cleaner to just have the `"`/`'` change in a single commit, but as I understand unless you're reassigning a variable, I think the `+=` and `++` have the same effect? I might be missing something though. I suppose to keep the same behaviour, the `++` should come before the argument, like `++$x` instead of `$x++`? (The difference being the new value is returned, vs the old value)

dstockto:
I'll undo the ++ changes-  The IDE suggested them as being more efficient. Since it's the reference code, the student doesn't see it anyway, so it doesn't really matter. The prefix ++ vs postfix ++ doesn't matter in this case as there's nothing else happening on the line so it really makes no difference. I have read before that prefix is slightly faster, but I don't know if that's ever true or if it was, if it still is. I personally prefer the postfix, but as you mentioned, it's personal preference, and I really don't care enough to keep it. I'll revert back to += later this afternoon and push up the changes.

dstockto:
All updated and rebased on top of the config.json fixes.

neenjaw:
Nice work @dstockto!

PR #428, opened 2021-10-03
Add union types concept

dstockto:


dstockto:
@neenjaw - Is this closer to what you're expecting? Future ones would be against empty concepts of course.

concepts/union-types/introduction.md
```diff
@@ -8,3 +8,71 @@ function add(int|float $a, int|float $b): int|float
     return $a + $b;
 }
 ```
+
+
+## Covariance and Contravariance
```
neenjaw, line 13: This is good content, but it reads like covariance and contravariance are almost a concepts unto themselves.  I don't think I have this one from that initial brainstormed list, but maybe we can add another concept to be able to use this content since it makes an important point.  Then when we create exercises, we can position this concept after type unions or extending classes or even after both?

neenjaw (commented):
Left one comment, this is great content.  Can we explode that content into a separate concept?

What do you think?

concepts/union-types/introduction.md
dstockto: I think that could make sense - I'll split it out.

neenjaw (approved):

PR #430, opened 2021-10-03
Add covariance/contravariance concept

dstockto:


neenjaw:
I'll add the config files to the branch, this pr is awesome though

concepts/covariance-and-contravariance/about.md
```diff
@@ -0,0 +1,5 @@
+# Introduction
+
+The concepts of covariance and contravariance in PHP are important as PHP enforces these rules are followed when extending classes or interfaces.
```
neenjaw, line 3: Is there a small or limited example we can pull in to the introduction to illustrate this at a small scale?

neenjaw (commented):
One question about the introduction, everything else is awesome

concepts/covariance-and-contravariance/about.md
dstockto: Would it be appropriate to put in maybe Calculator and InvalidCalc from the about.md? I'm not sure of a very brief way to introduce examples for this concept.

neenjaw: Yea, maybe can add them in a single code block: a parent class, a child class that is valid under the rules and a child class that is invalid under the rules.

Also, did the contents of the `introduction` and `about` get reversed? Introduction should be more brief, about should be more detailed

dstockto: It definitely looks like they got switched - I did have some weird merge conflicts but I thought I sanity checked too. I'll swap and get some examples into about.md

dstockto: Er, into introduction.md

neenjaw (approved):

Issue #414, opened 2021-09-19
Error when running php/resistor-color test


DavidLJz:
```
There was 1 error:

1) ResistorColorTest::testColors
Use of undefined constant COLORS - assumed 'COLORS' (this will throw an Error in a future version of PHP)

/home/user/exercism/php/resistor-color/ResistorColorTest.php:51
```
The error is present in the online editor too

neenjaw:
Can you paste your solution?

DavidLJz:
Sure

```php
<?php

declare(strict_types=1);

function colorCode(string $color): int
{
    $colors = [
        'black' => 0,
        'brown' => 1,
        'red' => 2,
        'orange' => 3,
        'yellow' => 4,
        'green' => 5,
        'blue' => 6,
        'violet' => 7,
        'grey' => 8,
        'white' => 9,
    ];

    return $colors[ strtolower($color) ];
}

```
All files have their original names.

neenjaw:
Ok cool, nice solution.  The error is originating because the test file expects you to define `COLORS` as a constant in the global scope (see the last test here: `https://github.com/exercism/php/blob/main/exercises/practice/resistor-color/ResistorColorTest.php`).  Since you are defining it within your function, the test case can't access it.

DavidLJz:
Oh I see. I think the instructions could be a bit more explicit then.

neenjaw:
Yes, at the very least it could be prudent to move it up to the first test case, since it would probably change the construction of a solution.

PR #435, opened 2021-10-09
Updated Resistor Color Exercise 

James-N-M:
Resistor color exercise error [Issue #414](https://github.com/exercism/php/issues/414)

Although nothings wrong with the exercise itself, it was mentioned in the comments that the last test should be moved to the first position, to help the students with their solution. Also I updated the instructions file to match a more up to date version from the js track.


ErikSchierboom:
We are still working on that @neenjaw. Sorry about the delay.

PR #436, opened 2021-10-11
Remove checking for the color array

AthulMuralidhar:
- i do not think that we need to check for the color array
- as in my solution, i defined the colour array inside the function itself

Here's a preview of my code:
```php

function colorCode(string $color): int
{
    $array = array(
"black" => 0,
"brown" => 1,
"red"=> 2,
"orange"=> 3,
"yellow"=> 4,
"green"=> 5,
"blue"=> 6,
"violet"=> 7,
"grey"=> 8,
"white"=> 9,
);
return $array[$color];
}
```


AthulMuralidhar:
I see, i didn't know that there was a spec like this haha. Just thought that checking for a const must me arbitrary, or at least mentioned somewhere so that the noob( me) knows to program it in. Thanks for the problem and keep up the good work ;)

PR #440, opened 2021-10-30
Update Isogram function name

meganos:
Unit test cases in [IsogramTest](https://github.com/exercism/php/blob/main/exercises/practice/isogram/IsogramTest.php) are looking for `isIsogram`  resulting in an undefined function error when running tests.

```
IsogramTest::testIsogram
Error: Call to undefined function isIsogram()

/mnt/exercism-iteration/IsogramTest.php:36
```

### DIFF
--- a/exercises/practice/isogram/Isogram.php
+++ b/exercises/practice/isogram/Isogram.php
@@ -24,7 +24,7 @@
 
 declare(strict_types=1);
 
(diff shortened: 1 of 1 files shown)


kotp (approved):
Thanks. Now it will match the names asked for in the test file.

PR #452, opened 2022-04-03
Update Matrix exercise to use real newlines

dstockto:
The description of this exercise talks about the data coming in with rows being separated by newlines, but in PHP, newlines are only newlines if they are in double quotes. Unfortunately, all the tests and the example solution are using the string '\n'. This PR updates the incoming string inputs to be double quoted so the newlines are actually newlines.

I've updated the tests and the example and it all appears to be working.

### DIFF
--- a/exercises/practice/matrix/.meta/example.php
+++ b/exercises/practice/matrix/.meta/example.php
@@ -47,7 +47,7 @@ class Matrix
 
     private function parseRows(string $matrix): void
     {
-        foreach (explode('\n', $matrix) as $stringRow) {
+        foreach (explode("\n", $matrix) as $stringRow) {
             $this->rows[] = array_map(fn(string $value) => (int) $value, explode(' ', $stringRow));
         }
     }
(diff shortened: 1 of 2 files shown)


neenjaw (approved):

neenjaw:
Good good, nice catch/update.  I think we can ignore the configlet warning for now, but I've asked @ErikSchierboom in slack. My concern is that because the config.json fails the configlet check, this fix will not get sync'd to the website.

dstockto:
It looks like it's expecting it to be listed as a slug in the config.json (along with a bunch of others). I am assuming if we add those then that starts making concepts show up on the PHP track, but we're probably not ready for that yet?

neenjaw:
Yea, I'm not sure if it will or not. In php's config.json we have concept exercises set to false, so maybe this isn't a concern. 

I think getting this fix merged is important though, would you mind adding those concepts to the config.json? If you are on mac/Linux running `uuidgen` in the terminal will generate a new UUID.

dstockto:
Yeah, I can do that - I guess I'd prefer to do that in another PR since it's unrelated to these changes. I can probably get to it a bit later this afternoon.

neenjaw:
Sure, ya, this one or a new one -- up to you.

dstockto:
I was looking at what this structure should look like based on the Go track - it looks like it's expecting an exercise for the concepts which I don't think we have.

dstockto:
I am able to get configlet lint to pass with this PR: https://github.com/exercism/php/pull/453

I am sure it will need to be updated when we get more into implementing the concepts... concept.

PR #467, opened 2022-06-06
`https` instead of `http`

NhanAZ:
I realized that some links are still using `http` protocol.
There are 28 links using `https` in 26 files and there are 10 links using `http` in 10 files.
This causes async? and `http` is not secure so I edited the whole link using `http` protocol to `https`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Exercism exercises in PHP
 - Follow the [contributing guide].
 - CI is run on all pull requests, it must pass the required checks for merge.
 
-[psr-12]: http://www.php-fig.org/psr/psr-12
+[psr-12]: https://www.php-fig.org/psr/psr-12
 [contributing guide]: https://github.com/exercism/x-api/blob/master/CONTRIBUTING.md#the-exercise-data
-[@group annotation]: http://phpunit.de/manual/4.1/en/appendixes.annotations.html#appendixes.annotations.group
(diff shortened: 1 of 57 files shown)


NhanAZ:
I just edited wikipedia's links. There are many other links that still use `http` but I'm not sure so I've ignored them.

NhanAZ:
If required I will edit the rest of the links.

neenjaw:
:+1: Would strongly recommend automating this process: `find ./exercises -type f -print0 | xargs -0 sed -i 's/http:\/\//https:\/\//g'`

NhanAZ:
What about links that require `http`?

neenjaw:
Do you have an example link that requires `http`?

NhanAZ:
Local links :thinking:

neenjaw:
When I run `grep -r -e 'http:\/\/'` I don't see any links that cannot be converted.  Am I missing something?

NhanAZ:
I mean some local links require http instead of https
It's still a bit risky to switch all http to https
Anyway, most browsers automatically convert http to https

neenjaw:
ah, yes, I understand now.  Yes, if there was a local web server interface that we referenced, yes it would require it to remain `http`.

neenjaw (approved):

NhanAZ:
Great it's merged, thanks for the review PR. Wish you a good day!

PR #471, opened 2022-06-29
Update TrinaryTest.php

razvbir:
My solution should not have passed all the tests. For 311 it gives 31, but the correct answer is 0

function toDecimal(string $number): int
{
    return array_reduce(
                str_split(strrev($number)),
                fn (array $acc,int $digit): array => array(
                    'power' => $acc['power'] + 1,
                    'valid' => $acc['valid'] && in_array($digit,range(0,2)),
                    'sum' => ($digit * 3 ** $acc['power'] + $acc['sum']) * $acc['valid'],
                ),
                array(
                    'power' => 0,
                    'sum' => 0,
                    'valid' => 1
                )
            )['sum'];
}

### DIFF
--- a/exercises/practice/trinary/TrinaryTest.php
+++ b/exercises/practice/trinary/TrinaryTest.php
@@ -75,4 +75,9 @@ class TrinaryTest extends PHPUnit\Framework\TestCase
     {
         $this->assertSame(0, toDecimal('13201'));
     }
+    
+    public function test311IsDecimal0(): void
+    {
+        $this->assertSame(0, toDecimal('311'));
+    }
 }


neenjaw:
This is a legacy exercise and has been deprecated.  So I don't think we should add this test case.  I would consider a PR to mark this exercise as such if you are interested.

To do this, we need to add the `"status": "deprecated"` to the exercise's entry in the `./config.json` file.

---

As a rule though, we ask that when people encounter missing test cases, it first be discussed as an issue in the problem-specifications repo (https://github.com/exercism/problem-specifications), so that it can be discussed and subsequently distributed to all of the different exercism tracks.

neenjaw:
I'm going to close this PR, buy as I commented previous if you'd like to open another with the deprecated status, I'll merge it.

PR #475, opened 2022-07-15
Update RobotNameTest.php

jasper2virtual:
```
    // the requirement say  "Your solution must ensure that every existing robot has a unique name."
    // but have not say not allow the same robot instance, reuse previous used name later.
    // so this test case is not valid
//     public function testNameArentRecycled(): void
//     {
//         $names = [];

//         for ($i = 0; $i < 10000; $i++) {
//             $name = $this->robot->getName();
//             $this->assertArrayNotHasKey($name, $names, sprintf('Name %s reissued after Reset.', $name));
//             $names[$name] = true;
//             $this->robot->reset();
//         }
//     }

    // This test is optional.
    public function testNameUniquenessManyRobots(): void
    {
        $names = [];
        $robots=[];
        for ($i = 0; $i < 10000; $i++) {
            $robot=new Robot();
            $name = $robot->getName();
            $this->assertArrayNotHasKey($name, $names, sprintf('Name %s reissued after %d robots', $name, $i));
            $names[$name] = true;
            // propose - keep the robot object, align to the requirement "Your solution must ensure that every existing robot has a unique name."
            // because the object will GC.
            array_push($robots,$robot);
        }
    }
```

### DIFF
--- a/exercises/practice/robot-name/RobotNameTest.php
+++ b/exercises/practice/robot-name/RobotNameTest.php
@@ -73,27 +73,34 @@ class RobotNameTest extends PHPUnit\Framework\TestCase
         $this->assertMatchesRegularExpression('/\w{2}\d{3}/', $name2);
     }
 
-    public function testNameArentRecycled(): void
-    {
-        $names = [];
+    // the requirement say  "Your solution must ensure that every existing robot has a unique name."
+    // but have not say not allow the same robot instance, reuse previous used name later.
+    // so this test case is not valid
+//     public function testNameArentRecycled(): void
+//     {
+//         $names = [];
 
-        for ($i = 0; $i < 10000; $i++) {
-            $name = $this->robot->getName();
-            $this->assertArrayNotHasKey($name, $names, sprintf('Name %s reissued after Reset.', $name));
-            $names[$name] = true;
-            $this->robot->reset();
-        }
-    }
+//         for ($i = 0; $i < 10000; $i++) {
+//             $name = $this->robot->getName();
+//             $this->assertArrayNotHasKey($name, $names, sprintf('Name %s reissued after Reset.', $name));
+//             $names[$name] = true;
+//             $this->robot->reset();
+//         }
+//     }
 
     // This test is optional.
     public function testNameUniquenessManyRobots(): void
     {
         $names = [];
-
+        $robots=[];
         for ($i = 0; $i < 10000; $i++) {
-            $name = (new Robot())->getName();
+            $robot=new Robot();
+            $name = $robot->getName();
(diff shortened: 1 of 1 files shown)


neenjaw:
Hello,

I think we can improve this test suite, but I don't agree with this change. Considering there are 676000 possible combinations of robot names (26 * 26 * 10 * 10 * 10), asking for someone to come up with a generator for 10000 names is not too much to ask for -- I don't think it is an important detail if the robots exist or not at this point.

If we want to make this test suite more exhaustive, I think we should look at `go`'s test suite for inspiration: https://github.com/exercism/go/blob/main/exercises/practice/robot-name/robot_name_test.go

I would be happy to explore alternative/additional test cases, but I think this one is fine.

exercises/practice/robot-name/RobotNameTest.php
```diff
@@ -73,27 +73,34 @@ public function testResetName(): void
...
+//     {
+//         $names = [];
+
+//         for ($i = 0; $i < 10000; $i++) {
+//             $name = $this->robot->getName();
+//             $this->assertArrayNotHasKey($name, $names, sprintf('Name %s reissued after Reset.', $name));
+//             $names[$name] = true;
+//             $this->robot->reset();
+//         }
+//     }
```
kotp, line 89: This should be not be a comment.  Code comments really do not belong, but the comment on line 76-78 could be a commit message.

exercises/practice/robot-name/RobotNameTest.php
```diff
@@ -73,27 +73,34 @@ public function testResetName(): void
...
+        $robots=[];
         for ($i = 0; $i < 10000; $i++) {
-            $name = (new Robot())->getName();
+            $robot=new Robot();
+            $name = $robot->getName();
             $this->assertArrayNotHasKey($name, $names, sprintf('Name %s reissued after %d robots', $name, $i));
             $names[$name] = true;
+            // propose - keep the robot object, align to the requirement "Your solution must ensure that every existing robot has a unique name."
+            // because the object will GC.
+            array_push($robots,$robot);
```
kotp, line 103: Same for this comment.  It should be a discussion in the issue, the pull request, or a commit message, but not here.

kotp (changes requested):
This said, the exercise is about optimization, and we should be able to generate all names and ensure no duplicates.  Otherwise the exercise does not honor the README.md part of the specification.

Notice the "must" wording:  "Your solution must ensure that every existing robot has a unique name."

So that is not an optional condition of the exercise.

neenjaw:
Thank @kotp for the discussion points. I think though that making requests on the existing point about the comments doesn't help guide it toward a mergable solution.

@jasper2virtual, I'm going to close this PR, please open an issue to discuss this further, or if you want to open a new pr with the exhaustive test cases from the go track, that works too.

PR #481, opened 2022-08-27
Update  pascalsTriangleRows to allow nullable type

meganos:
The current version of `pascalTriangleRows` has an `int` type declaration for its parameter `$rowCount`, but this causes one of the unit tests to fail:

```
There was 1 error:

1) PascalsTriangleTest::testNullNoRows
TypeError: pascalsTriangleRows(): Argument #1 ($rowCount) must be of type int, null given
```

Marking it as nullable results in all unit tests passing.


meganos:
On second thought, this proposed change is unnecessary. Current approach allows student to explore nullable type hints and add the change.

PR #511, opened 2023-06-01
[Fix] testColors to green

amarhaj:
One of Resistor Color tests is failing - testColors. By other tests seems that assoc arrray is meant to be used here. Updated test to green.


neenjaw:
Can you give me some context for why you want to make this change? We run continuous integration on the exercises in this repository, and this previously hasn't been a documented problem.

```
Running tests for: resistor-color
PHPUnit 9.6.8 by Sebastian Bergmann and contributors.

....                                                                4 / 4 (100%)

Time: 00:00.001, Memory: 24.41 MB

OK (4 tests, 4 assertions)
```
> https://github.com/exercism/php/actions/runs/4989720003/jobs/8934011069

PR #531, opened 2023-08-01
[fix] update pizza pi instructions for excersise number one

HenryVogt:
Currently the instructions are misleading. The description says that for every person that is served, 45g of dough are added. The example below calculates with 20g. As a result the corresponding test fails because the expected amount of dough is only true with 20g per person.


neenjaw (approved):
Thanks for the typo fix! 🎉

PR #534, opened 2023-08-03
Add the Bank Account exercise

MichaelBunker:
Added the Bank Account exercise without the concurrency test.

See: https://github.com/exercism/problem-specifications/tree/main/exercises/bank-account

PR #538, opened 2023-08-05
Add the resistor-color-duo exercise.

MichaelBunker:
See the specification: https://github.com/exercism/problem-specifications/tree/main/exercises/resistor-color-duo

### DIFF
--- a/exercises/practice/resistor-color-duo/.meta/example.php
+++ b/exercises/practice/resistor-color-duo/.meta/example.php
@@ -24,12 +24,24 @@
 
 declare(strict_types=1);
 
-function getAllColors(): array
+class ResistorColorDuo
 {
-    throw new \BadMethodCallException("Implement the getAllColors function");
-}
+    public function getAllColors(): array
+    {
+        return ["black", "brown", "red", "orange", "yellow", "green", "blue", "violet", "grey", "white"];
+    }
 
-function colorCode(string $color): int
-{
-    throw new \BadMethodCallException("Implement the colorCode function");
+    /**
+     * @param string[] $colors
+     */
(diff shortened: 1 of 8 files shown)


exercises/practice/resistor-color-duo/.meta/example.php
```diff
@@ -0,0 +1,42 @@
...
+ * <link>.
+ *
+ * To disable strict typing, comment out the directive below.
+ */
+
+declare(strict_types=1);
+
+class ResistorColorDuo
+{
+    public const COLORS = ["black", "brown", "red", "orange", "yellow", "green", "blue", "violet", "grey", "white"];
```
neenjaw, line 29: I think this is the right implementation as it is built from the last exercise, could you look at adding some hints to the previous exercise for students to use a constant here? Or maybe we shouldn't use a public array constant at all, and specify another method to list the colors to make the implementation and learning goals more obvious?

neenjaw (approved):
Looks good, once you look at the feedback for https://github.com/exercism/php/pull/534 I think a few of the same comments can be applied here, but I don't want to nag.

Happy to merge this optimistically, let me know what you'd like to do

exercises/practice/resistor-color-duo/.meta/example.php
MichaelBunker: I changed it to a method to make it more obvious. seeing just `COLORS` in the test file might be confusing for someone just starting out.

MichaelBunker:
@neenjaw I've made changes that you requested to resistor-color and resistor-color-duo. This is ready for review.

neenjaw (approved):

PR #541, opened 2023-08-06
Add the resistor-color-trio exercise.

MichaelBunker:
See specification: https://github.com/exercism/problem-specifications/tree/main/exercises/resistor-color-trio

### DIFF
--- a/exercises/practice/resistor-color-trio/.meta/example.php
+++ b/exercises/practice/resistor-color-trio/.meta/example.php
@@ -1,0 +1,85 @@
+<?php
+
+/*
+ * By adding type hints and enabling strict type checking, code can become
+ * easier to read, self-documenting and reduce the number of potential bugs.
+ * By default, type declarations are non-strict, which means they will attempt
+ * to change the original type to match the type specified by the
+ * type-declaration.
+ *
+ * In other words, if you pass a string to a function requiring a float,
+ * it will attempt to convert the string value to a float.
+ *
+ * To enable strict mode, a single declare directive must be placed at the top
+ * of the file.
+ * This means that the strictness of typing is configured on a per-file basis.
+ * This directive not only affects the type declarations of parameters, but also
+ * a function's return type.
+ *
+ * For more info review the Concept on strict type checking in the PHP track
+ * <link>.
+ *
+ * To disable strict typing, comment out the directive below.
+ */
+
+declare(strict_types=1);
+
+class ResistorColorTrio
+{
+    public const GIGA_BREAKPOINT = 1000000000;
+    public const MEGA_BREAKPOINT = 1000000;
+    public const KILO_BREAKPOINT = 1000;
+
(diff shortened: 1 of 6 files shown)


exercises/practice/resistor-color-trio/.meta/example.php
```diff
@@ -0,0 +1,81 @@
...
+declare(strict_types=1);
+
+class ResistorColorTrio
+{
+    public const GIGA_BREAKPOINT = 1000000000;
+    public const MEGA_BREAKPOINT = 1000000;
+    public const KILO_BREAKPOINT = 1000;
+    public const COLORS = ['black', 'brown', 'red', 'orange', 'yellow', 'green', 'blue', 'violet', 'grey', 'white'];
+
+    public static function label(array $colors): string
```
neenjaw, line 34: Out of curiosity, why did you choose a static method implementation here?

neenjaw (approved):
Looks good, once you look at the feedback for https://github.com/exercism/php/pull/534 I think a few of the same comments can be applied here, but I don't want to nag.

Happy to merge this optimistically, let me know what you'd like to do

exercises/practice/resistor-color-trio/.meta/example.php
MichaelBunker: I keep going back and forth on to use static or not for exercises. There are some good use cases for using the static keyword in PHP, but I'm not sure they would apply in these exercises. Since many of the functions are pure functions, having to instantiate the class doesn't seem strictly necessary, especially if the purpose of the exercise isn't to teach about classes and/or OOP specifically. I'm not sure if the static keyword helps or hinders students who are learning.

Does the PHP track have a preference? It wasn't clear from the existing exercises but I can just as easily avoid the static keyword if that's preferable.

neenjaw: So far, my preference has been not to use it, but I don't think there's anything scientific about that choice.  My take on this in PHP I believe is guided by these things:
- We don't really have formal import/export properties to lean on to organize code
- Unit testing in PHP (like with PHP unit) is primarily oriented around instance objects (allowing you to swap them out with mocks or via interface).
- `static` is _one more_ keyword (that eventually must be learned to master the language -- no doubt about that) and that does increase the complexity
  - it also has subsequent design considerations since static methods can only access static properties, etc

I'm happy to defer to your preference here, but if this is an exercise geared at an "early" PHP learner, then maybe a class instance is easier to work with and has fewer sharp edges.

MichaelBunker: Those are good points. I'll change to instantiate the class instead of using static methods for this and my other open PRs.

MichaelBunker:
@neenjaw I've made the changes you requested. This PR is ready for review.

neenjaw (approved):
Really nice work here, really appreciate this

PR #540, opened 2023-08-06
Add the twelve-days exercise.

MichaelBunker:
Opened a new PR since I rebased to fix an issue while the first was closed and it couldn't be reopened. 

See specification: https://github.com/exercism/problem-specifications/tree/main/exercises/twelve-days

PR #547, opened 2023-08-20
Add alphametics exercise and update TwelveDays stub.

MichaelBunker:
### Add the alphametics exercise.

See the problem specification here: https://github.com/exercism/problem-specifications/blob/main/exercises/alphametics/canonical-data.json

I referenced the Go and JS repos for the example solution here.

Also removed the static keyword from the stub for twelve-days exercise. The tests and example were updated to not use static keywords but the stub was missed in https://github.com/exercism/php/pull/540

### DIFF
--- a/config.json
+++ b/config.json
@@ -885,6 +885,14 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 3
+      },
+      {
+        "slug": "alphametics",
+        "name": "Alphametics",
+        "uuid": "49d212c6-6fa3-4e7b-95d9-595613630566",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 5
       }
     ]
   },
--- a/exercises/practice/alphametics/.docs/instructions.md
+++ b/exercises/practice/alphametics/.docs/instructions.md
@@ -1,0 +1,31 @@
+# Instructions
+
+Write a function to solve alphametics puzzles.
+
+[Alphametics][alphametics] is a puzzle where letters in words are replaced with numbers.
+
+For example `SEND + MORE = MONEY`:
+
+```text
+  S E N D
+  M O R E +
+-----------
+M O N E Y
+```
+
+Replacing these with valid numbers gives:
+
+```text
+  9 5 6 7
+  1 0 8 5 +
+-----------
+1 0 6 5 2
+```
+
+This is correct because every letter is replaced by a different number and the words, translated into numbers, then make a valid sum.
+
+Each letter must represent a different digit, and the leading digit of a multi-digit number must not be zero.
+
+Write a function to solve alphametics puzzles.
+
+[alphametics]: https://en.wikipedia.org/wiki/Alphametics
--- a/exercises/practice/alphametics/.meta/config.json
+++ b/exercises/practice/alphametics/.meta/config.json
@@ -1,0 +1,10 @@
+{
+  "blurb": "Write a function to solve alphametics puzzles.",
+  "authors": ["MichaelBunker"],
+  "contributors": [],
+  "files": {
+    "solution": ["Alphametics.php"],
+    "test": ["AlphameticsTest.php"],
+    "example": [".meta/example.php"]
+  }
+}
--- a/exercises/practice/alphametics/.meta/example.php
+++ b/exercises/practice/alphametics/.meta/example.php
@@ -1,0 +1,128 @@
+<?php
+
+/*
+ * By adding type hints and enabling strict type checking, code can become
+ * easier to read, self-documenting and reduce the number of potential bugs.
+ * By default, type declarations are non-strict, which means they will attempt
+ * to change the original type to match the type specified by the
+ * type-declaration.
+ *
+ * In other words, if you pass a string to a function requiring a float,
+ * it will attempt to convert the string value to a float.
+ *
+ * To enable strict mode, a single declare directive must be placed at the top
+ * of the file.
+ * This means that the strictness of typing is configured on a per-file basis.
+ * This directive not only affects the type declarations of parameters, but also
+ * a function's return type.
+ *
+ * For more info review the Concept on strict type checking in the PHP track
+ * <link>.
+ *
+ * To disable strict typing, comment out the directive below.
+ */
+
+declare(strict_types=1);
+
+class Alphametics
+{
+    public function solve(string $puzzle): ?array
+    {
+        //Remove operators and filter down to just the letters
+        $parts = preg_split('/[+|==]/', $puzzle);
+        $parts = array_map('trim', $parts);
+        $parts = array_filter($parts);
+
+        $firstLetters = [];
+
+        foreach ($parts as $part) {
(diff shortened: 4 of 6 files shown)


MichaelBunker:
> My only concern with this exercise is that some of the test cases are very slow. They will never finish using the online test-runner for most depth first search approaches within the 10s window.

### TLDR
I think something strange is happening when Ubuntu is running tests causing it to be much slower for these tests compared to the other OSes, and I'm not sure any improvements to the example code would be enough to get us under the threshold you've set. We could skip tests using groups in PHPUnit but that seems like an awkward workaround that would create a separate experience for users depending on if they are coding locally or live in Exercism.

### Slow Tests
I was a bit surprised that speed was an issue because the tests ran in under 10 seconds locally for me every time. Interestingly, the tests only seem excruciatingly slow on the linux boxes in the CI. Windows and MacOS seem to be under 10 seconds consistently. This also matches the results I see locally, (MacOS, PHPUnit 9, PHP 8.1). I don't know why this is, but that's an avenue to investigate I would think. I'm not sure what the possible issue is at first glance that is causing this.

```
Ubuntu
PHP 8
Running tests for: alphametics
Time: 00:47.996, Memory: 22.43 MB

PHP 8.1
Running tests for: alphametics
Time: 00:54.440, Memory: 22.43 MB

PHP 8.2
Running tests for: alphametics
Time: 00:46.738, Memory: 22.43 MB

Windows --
PHP 8
Running tests for: alphametics
Time: 00:07.339, Memory: 24.00 MB

PHP 8.1
Running tests for: alphametics
Time: 00:09.748, Memory: 24.00 MB

PHP 8.2
Running tests for: alphametics
Time: 00:10.363, Memory: 24.00 MB

MacOS -- 
PHP 8
Running tests for: alphametics
Time: 00:09.500, Memory: 22.43 MB

PHP 8.1
Running tests for: alphametics
Time: 00:08.571, Memory: 22.43 MB

PHP 8.2
Running tests for: alphametics
Time: 00:08.984, Memory: 22.43 MB
```

### Make the solution better
I could try and make the solution faster, there are some peculiarities with Alphametics that could maybe reduce the workload. I did spend some time with some shortcuts during the initial solution but ran into various problems.
Example:
The max value for any digit is 9, and the max possible sum of 9 + 9 (the S + M columns) is 18. So the carry over M must be 1. 
```
   S E N D
+  M O R E
----------
 M O N E Y
```
The wiki page has some info on some of these sort of deductions that might help. I had tried incorporating some, which helped marginally with some test cases, but not all of them.

**The problem is I'm not sure they would help with the more complicated test cases and I don't think they would solve the issue with Ubuntu running a lot slower. Even if I could shave off some time for some of the tests, Ubuntu would still be over 10 seconds given how long it's running compared to the other OSes.** 

### On Skipping Tests
To skip the tests we could add phpunit group annotations to the tests that are troublesome.
`@large` `@medium` `@small` will cause the tests to fail when they take over a certain threshold. It could also be any arbitrary group annotation that we skip in the CI by changing the command to exclude that group or updating the phpunit config file to do the same.

However it does seem strange to skip tests like that just on a theoretically level. If we have tests that have to be skipped, why include them at all? If users can't run tests in the live test-runner, it seems like it should just not be included at all.  It also seems like an awkward workaround that would create a 'live test runner' course, and a 'local test runner' course where users would have different concerns based on if they are using the live test runner or not. Which maybe that's ok or even encouraged by Exercism, but I had an assumption that the point of the live editor was to remove the barrier to entry, not change the experience itself. I.E., a person coding locally and a person coding with the live editor should be solving the same problems and doing the same work.

MichaelBunker:
I've poked around a little more with the php-test-runner repository and the CI results. I was expecting to start up a minimal php-alpine instance in Docker and have tests take a long time (I.E. same result as the CI) but the tests run fairly quickly which is not what I was expecting seeing the CI results.

It could be something related to the Exercism specific php build for tests, or maybe the CI jobs themselves, maybe I'm just missing something else obvious here. I'll keep thinking on it.

I copied `phpunit.phar`, `Alphametics.php` and `AlphameticsTest.php` from this branch to a new test directory with this Dockerfile and ran phpunit on the tests and it finished in under 4 seconds.

Dockerfile
```
FROM php:8.2.7-cli-alpine3.18

RUN adduser -D bunker
ADD alphametics /alphametics
RUN chmod -x /alphametics/phpunit-9.phar
RUN chown -R bunker:bunker /alphametics

USER bunker
```

Results
```sh
/ $ cat /etc/alpine-release 
3.18.2

/ $ whoami
bunker

/ $ cd alphametics/

/alphametics $ php -v
PHP 8.2.7 (cli) (built: Jun 15 2023 03:31:57) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.7, Copyright (c) Zend Technologies

/alphametics $ php ./phpunit-9.phar --version
PHPUnit 9.6.10 by Sebastian Bergmann and contributors.

/alphametics $ php ./phpunit-9.phar AlphameticsTest.php 
PHPUnit 9.6.10 by Sebastian Bergmann and contributors.
..........                                                        10 / 10 (100%)
Time: 00:03.227, Memory: 22.41 MB
OK (10 tests, 10 assertions)
```

neenjaw:
That's strange, yea, this ci action uses `shivammathur/setup-php` to set up php for the ci environment.  I wonder if there is a problem specific to that github action.

Thanks for the testing in alpine, it's quite reassuring.  I think we should move forward with this as it is and monitor for complaints.

MichaelBunker:
> That's strange, yea, this ci action uses `shivammathur/setup-php` to set up php for the ci environment. I wonder if there is a problem specific to that github action.
> 
> Thanks for the testing in alpine, it's quite reassuring. I think we should move forward with this as it is and monitor for complaints.
> 
> What do you think?

My concern would be if the online test runner is the same image that the CI job uses, then this problem will exist for the online test runner as well even though alpine works fine. The php-test-runner repo has the docker image for running tests, is that used for exercism whenever php tests are ran or is that specifically for the CI jobs?

I can look into the `shivammathur/setup-php` a bit and see if this issue is already discussed somewhere or if there is an obvious cause.

neenjaw:
The test runner for the website runs on an alpine image, if the concern was isolated to Ubuntu, I don't think there's much risk.

MichaelBunker:
Ok, well the solution is in a good spot so if you don't think it will be an issue for the online runner I have no issue with it.

I'm not familiar with how long it takes for an exercise to become visible on the website or if there is a "staging" site that this could be tested on, but perhaps we should put the example solution in a submission via the online editor and test the online test runner once the exercise is up. A quicker feedback loop doing that than waiting for an user to stumble on the potential problem of test length at some date in the future.

Is there a way to test the online runner before this hits the live website?

MichaelBunker:
@neenjaw I poked around a bit more with `shivammathur/setup-php` but don't see anything specific. I'm fine with merging and monitoring as you suggested. Let me know if you need anything else for this PR.

neenjaw:
Sounds good, I'll merge it and we can go from there.

PR #548, opened 2023-08-24
fix: several typos in concepts

vnkmpf:


neenjaw:
you need to fix also the underlying concepts which are used in the template generation.  you can find them in `/concepts`

neenjaw (changes requested):
see comment

vnkmpf:
Thanks for the comment.
I fixed some more typos in the directory.

There is still the mysterious failure of one check, even though the other PR worked :shrug:

neenjaw:
That CI failing is because that step generates a difference between the hydrated template and the resulting introduction document. So when it tries to commit the difference in the branch, it fails since it doesn't have permission to commit to your branch (nothing for you to fix there).  If you pull this locally then run:

```shell
./bin/fetch_configlet
```

then

```shell
configlet generate
```

you will be able to see the difference generated. So next time it might be easier to have this workflow:

1. Make a new branch locally
2. Make the changes to the introduction/about files in `/concepts`
3. run `config generate`
4. Then make the changes in the instructions and hints since they are not auto-generated.

(Side note, @ee7 can we add a comment to files computed by configlet generate that they are auto-generated and not to be manually edited? It might ease this process next time someone is onboarding)

vnkmpf:
@neejaw Thank you for posting those steps. I thought the problem was the same as before.
Now that I followed the steps, it seems to be fixed.

neenjaw (approved):
Thanks for the changes! Yes, maybe we could add those suggestion to that workflow when it fails to be a bit more helpful when this situation happens. Do you want to look into that process with GitHub actions?

vnkmpf:
I have zero experience with GitHub Actions, so it would be probably more of exploring and getting familiar with than help.
If it's something that is desired sooner than later, I think I cannot help much.
But if it's a nice-to-have, I might take a look :)

neenjaw:
It's just a nice to have, no real rush on it. :D

PR #550, opened 2023-08-26
Add the Binary Search Tree exercise.

MichaelBunker:
Problem Specification located here: https://github.com/exercism/problem-specifications/tree/main/exercises/binary-search-tree

### DIFF
--- a/exercises/practice/binary-search-tree/.meta/example.php
+++ b/exercises/practice/binary-search-tree/.meta/example.php
@@ -1,0 +1,71 @@
+<?php
+
+/*
+ * By adding type hints and enabling strict type checking, code can become
+ * easier to read, self-documenting and reduce the number of potential bugs.
+ * By default, type declarations are non-strict, which means they will attempt
+ * to change the original type to match the type specified by the
+ * type-declaration.
+ *
+ * In other words, if you pass a string to a function requiring a float,
+ * it will attempt to convert the string value to a float.
+ *
+ * To enable strict mode, a single declare directive must be placed at the top
+ * of the file.
+ * This means that the strictness of typing is configured on a per-file basis.
+ * This directive not only affects the type declarations of parameters, but also
+ * a function's return type.
+ *
+ * For more info review the Concept on strict type checking in the PHP track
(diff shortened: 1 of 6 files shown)


exercises/practice/binary-search-tree/.meta/example.php
```diff
@@ -0,0 +1,73 @@
...
+            $data = array_merge($data, $this->left->getSortedData());
+        }
+
+        if ($this->right) {
+            $data = array_merge($data, $this->right->getSortedData());
+        }
+
+        sort($data);
+
+        return $data;
```
neenjaw, line 61: Doesn't this result in an O(n*log(n)) complexity rather than an O(n) complexity because of the extra sort?

Can we do an getSorted more like https://medium.com/the-andela-way/binary-tree-implementation-in-php-e12df09d046f as the sorted values should be just an in-order traversal of the tree

I do like how you have this very simple that each node is an instance of the "Tree" class, I think that's good to keep to minimize the expectations of the exercise.

neenjaw (changes requested):
1 thought about the get sorted method, otherwise I think it looks great

exercises/practice/binary-search-tree/.meta/example.php
MichaelBunker: Thanks for the feedback. It looks like the `getSortedData()` function can be done without the last `sort()` on line 59.

The following method passes tests without the extra sort,  is this more in line with what you'd like to see? If not I can make it match more closely to how the article you posted is retrieving nodes.

```php
    public function getSortedData(): array
    {
        $data[] = $this->data;

        if ($this->left) {
            $data = array_merge($this->left->getSortedData(), $data);
        }

        if ($this->right) {
            $data = array_merge($data, $this->right->getSortedData());
        }

        return $data;
    }
```

neenjaw: Yea, that's a good reduction. I'd be good to accept that. Thanks for iterating on that!

neenjaw (approved):

PR #557, opened 2023-09-09
Update instructions.md

alastairsounds:
Fixed typo from
"The method should take in a string work"
to accurately reference argument $word
"The method should take in a string word"


neenjaw (approved):
Thanks for this typo fix!

PR #576, opened 2023-11-07
Add constructor template in LargestSeriesProduct.php

realwahl:
The tests imply that the Series class has a constructor with a string parameter. If we provide a class template, we should provide a template for the constructor as well.

### DIFF
--- a/exercises/practice/largest-series-product/LargestSeriesProduct.php
+++ b/exercises/practice/largest-series-product/LargestSeriesProduct.php
@@ -27,7 +27,7 @@ declare(strict_types=1);
 class Series
 {
     public function __construct(string $input)
-	{
+    {
         throw new \BadMethodCallException("Implement the constructor method");
     }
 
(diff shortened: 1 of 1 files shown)


mk-mxp:
@realwahl Are you willing to make this mergable?

homersimpsons:
@realwahl Tests are failing because a `[tab]` is used on line 30 instead of 4 spaces. Do you have time to fix this ?

exercises/practice/largest-series-product/LargestSeriesProduct.php
```diff
@@ -26,6 +26,11 @@
 
 class Series
 {
+    public function __construct(string $input)
+	{
```
homersimpsons, line 30: Replace tab by 4 spaces:
```suggestion
    {
```

homersimpsons (changes requested):

realwahl:
I will check, thanks!

realwahl:
@mk-mxp  I think it should be ok now, can you please check?
@homersimpsons Thanks a lot!

homersimpsons (approved):

PR #577, opened 2023-11-07
Add requested methods to Matrix.php template

realwahl:
When offering a template, it should contain the expected signatures of the methods.

### DIFF
--- a/exercises/practice/matrix/Matrix.php
+++ b/exercises/practice/matrix/Matrix.php
@@ -26,8 +26,18 @@ declare(strict_types=1);
 
 class Matrix
 {
-    public function __construct()
+    public function __construct(string $matrix)
     {
         throw new BadFunctionCallException("Please implement the Matrix class!");
+    }
+
+    public function getRow(int $rowId): array
+    {
+        throw new BadFunctionCallException("Please implement the getRow method!");
+    }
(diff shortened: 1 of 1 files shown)


neenjaw:
> When offering a template, it should contain the expected signatures of the methods.

The decision to offer templates and method stubs for some exercises is a relatively new considering the age of the PHP exercises. Appreciate the addition here, however since one of exercism's tenants is to teach test driven development I think we should balance this with maybe looking into how we can redirect a person to look at the test file for practice exercises

neenjaw:
I'll have to see what's going on with CI here, may be either affected by a github CI issue or perhaps something has changed in how the Ci builds PHP for the tests to be run

neenjaw:
> we either provide all correct stubs or none (with a comment explaining what we want from the student)

@realwahl, sure that can be our target. In this one give all the stubs, in others if there is an incorrect constructor which doesn't match the test expectation we can fix that and/or provide a comment

For this PR, I think we need to figure out what is crashing CI, such that it can't complete -- configlet is a required CI step.   Is it possible for you to merge or rebase on the newest master revision?

neenjaw (approved):
✅ but I can't merge it until we get CI passing

PR #581, opened 2023-11-11
Concept of type juggling

mk-mxp:
I read into the `type juggling` concept and found it to be a bit to concise. Here's my suggestions to:

* clarify what PHP does when coersion happens
* emphasize the type changes in the examples
* make clear, that coersion does not alter the variable value
* show `strict_types` to turn off type coercion

I also have questions about the whole mechanism behind the concepts and the exercises:

* The `/concepts/*/about.md` are the texts shown on `https://exercism.org/tracks/php/concepts/type-juggling`. Any other `*.md` in this directory can be imported into `/exercises/concept/*/.docs/introduction.md.tpl`? Or just `introduction.md`?
* How to convert `/exercises/concept/*/.docs/introduction.md.tpl` to the actual file? Is that happening magically for me?
* How do I tell exercism, that I have more than one exercise for a concept? Is that all taken from the exercise list in  `/config.json`?

### DIFF
--- a/concepts/type-juggling/about.md
+++ b/concepts/type-juggling/about.md
@@ -1,6 +1,7 @@
 # Type Juggling
 
-Type juggling may also be known as type coercion. Type juggling is when two values of different types are coerced to the same type to perform an operation.
+Type juggling may also be known as type coercion.
+Type juggling is when two values of different types are coerced to the same type to perform an operation.
 
 ```php
 $baskets = 5 // int
@@ -9,9 +10,11 @@ $baskets * $apples_per_basket
 # => 15 // int
 ```
 
-This implicitly converts the `string` to an `int` to execute the `*` operation. The decision, which type to coerce, is made by PHP according to the rules of the operation and these rules vary largely.
+This implicitly converts the `string` to an `int` to execute the `*` operation.
+The decision, which type to coerce, is made by PHP according to the rules of the operation and these rules vary largely.
 
-Type juggling **does not change** the type of the value stored in a variable. Only the value processed by the operation is converted.
+Type juggling **does not change** the type of the value stored in a variable.
+Only the value processed by the operation is converted.
 
 ```php
 $apples_per_basket = "3" // string
@@ -28,7 +31,8 @@ strlen(12321);
 
 ## Turn off type juggling
 
-Type juggling happens by default. We can stop PHP from coercing and force it to check types strictly:
+Type juggling happens by default.
+We can stop PHP from coercing and force it to check types strictly:
 
 ```php
 <?php
@@ -48,9 +52,11 @@ $my_number = (int) $apples_per_basket // cast string to int
 # => 3 // int
 ```
 
-This will forcibly convert the value to the type specified in the brackets `(...)`. Allowed types for manual type casting are: `bool`, `int`, `float`, `string`, `array`, `object`.
+This will forcibly convert the value to the type specified in the brackets `(...)`.
+Allowed types for manual type casting are: `bool`, `int`, `float`, `string`, `array`, `object`.
 
-Type casting **does not change** the type of the value stored in a variable. Only the value processed by the operation is converted.
+Type casting **does not change** the type of the value stored in a variable.
+Only the value processed by the operation is converted.
 
 ```php
 $apples_per_basket = "3" // string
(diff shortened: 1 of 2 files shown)


concepts/type-juggling/about.md
```diff
@@ -1,25 +1,76 @@
...
+
+```php
+strlen(12321);
+# => 5
+```
+
+## Turn off type juggling
+
+Type juggling happens by default.
+We can stop PHP from coercing and force it to check types strictly:
```
neenjaw, line 35: This isn't accurate -- https://www.php.net/manual/en/language.types.declarations.php

This declaration only disables coercion on function parameters and return types made by calls from the file where the declaration is made. 

It doesn't turn off type juggling globally or in every situation.

I think this is worth having its own separate concept.

concepts/type-juggling/about.md
```diff
@@ -1,25 +1,76 @@
...
 $apples_per_basket = "3" // string
 $baskets * $apples_per_basket
-# => 15
+# => 15 // int
+```
+
+This implicitly converts the `string` to an `int` to execute the `*` operation.
+The decision, which type to coerce, is made by PHP according to the rules of the operation and these rules vary largely.
+
+Type juggling **does not change** the type of the value stored in a variable.
```
neenjaw, line 16: Is this a necessary detail? If we add this then I think you almost need to add it to every operation that isn't explicitly assigning the value to a variable, which feels redundant as we repeat this statement twice in this description.

If there is a particular point of concern here about how variable assignment occurs, we should address that in the variable assignment concept.

concepts/type-juggling/about.md
```diff
@@ -1,25 +1,76 @@
...
 ```php
-$my_number = "3"
-(int) 3
-# => 3
+$apples_per_basket = "3" // string
+$my_number = (int) $apples_per_basket // cast string to int
+# => 3 // int
+```
+
+This will forcibly convert the value to the type specified in the brackets `(...)`.
```
neenjaw, line 55: Can we drop "forcibly" from this statement?

neenjaw:
For your question they would be good to ask on the forum. I know @iHiD is planning to make some changes here about how/when concept information is shown to the student, so would be good to get that from him and have it updated in the official build documentation. 

I also don't know what is available for use in the templating system, another good question for the forum to be clarified.

concepts/type-juggling/about.md
```diff
@@ -1,25 +1,76 @@
...
 
 ```php
 $baskets = 5 // int
 $apples_per_basket = "3" // string
 $baskets * $apples_per_basket
-# => 15
+# => 15 // int
+```
+
+This implicitly converts the `string` to an `int` to execute the `*` operation.
```
neenjaw, line 13: Rather than use "this" or "it" can we be clear about the mechanisms in action?

```suggestion
When the expression is evaluated, the string value is implicitly converted to an `int` to match the expectation of the `*` operator.
```

concepts/type-juggling/about.md
```diff
@@ -1,25 +1,76 @@
...
-$my_number = "3"
-(int) 3
-# => 3
+$apples_per_basket = "3" // string
+$my_number = (int) $apples_per_basket // cast string to int
+# => 3 // int
+```
+
+This will forcibly convert the value to the type specified in the brackets `(...)`.
+Allowed types for manual type casting are: `bool`, `int`, `float`, `string`, `array`, `object`.
```
neenjaw, line 56:
```suggestion
All primitive types can be used can be used as the target of type. `bool`, `int`, `float`, `string`, `array`, `object`.
```

concepts/type-juggling/about.md
mk-mxp: Well, there's always something to learn. Thanks for pointing that out to me!

Strict types will get their own concept for sure...

concepts/type-juggling/about.md
mk-mxp: It was one of the most asked questions on this topic I was facing. Often implicit, when people tried to show the conversion result after the operation by `var_dump()`-ing the variable.

So I think, yes it is necessary detail. I gave it its own heading, so it is not repeated. And I used an example without assignment.

neenjaw: This sounds like a good approach if you've seen it come up in reviews

neenjaw (approved):

PR #585, opened 2023-11-17
Deprecate scale-generator

BNAndras:
See https://forum.exercism.org/t/scale-generator-tests/8236/1


neenjaw:
@BNAndras I'm not sure what to do with this yet.  I can see that it can be a hard toy problem for some students, and it isn't in the best shape, but I also think that to deprecate means invalidating a lot of solutions that have been done for this problem.

mk-mxp:
As [shown in the forum](https://forum.exercism.org/t/soft-remove-deprecated-exercises/10903/2) and described in the [documentation](https://exercism.org/docs/building/tracks/deprecated-exercises), there is no invalidation of solutions (if that means they are lost for the students or students loose reputation earned). The exercise will be hidden for new students only.

If there is no other reason not to deprecate this exercise, I suggest: Let's get that merged!

mk-mxp:
@BNAndras I can't rebase this branch to fix the formatting issues. Do you want to rebase your branch or should I close this and deprecate the exercise in a new PR?

PR #587, opened 2023-11-27
fix: explode with empty separator in exemplar

vnkmpf:
Explode with empty separator errs: https://3v4l.org/RevTj
Use mb_str_split instead: https://3v4l.org/VAKbR

### DIFF
--- a/exercises/concept/city-office/.meta/exemplar/Form.php
+++ b/exercises/concept/city-office/.meta/exemplar/Form.php
@@ -9,7 +9,7 @@ class Form
 
     function letters(string $word): array
     {
-        return explode("", $word);
+        return mb_str_split($word);
     }
 
     function checkLength(string $word, int $max_length): bool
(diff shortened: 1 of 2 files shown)


neenjaw:
I need a bit more of an explanation.  Why does CI pass using the exemplar right now:? [CI Logs](https://github.com/exercism/php/actions/runs/7012366068/job/19076749031)

vnkmpf:
Maybe I didn't express myself clearly, why I think it's a problem.

The exercise wants you to provide types, so the tests don't test the actual behavior, which is provided in the file, but only tests correct types are added.

Why I think it should be fixed is that it shows you a code that would not work. As seen in 3v4l examples.
Taken to a bit extreme example - this code would still pass tests, but doesn't "feel right".
```
function letters(string $word): array
{
    throw new \RuntimeException();
}
```

neenjaw:
Makes sense

neenjaw (approved):

PR #595, opened 2024-01-12
:construction_worker: Run lint only once

homersimpsons:
Linting results will not depend of the OS nor of the PHP version. Let's save some CI time and the planet by running it only once.


neenjaw (approved):
considering how infrequently this CI runs, I don't think this is a priority concern. But you've already done the work so I am agreeable to approve.

Issue #578, opened 2023-11-07
Improve syllabus


mk-mxp:
Hello maintainers 👋

I program in PHP for ~15yrs now and would like to add my share to the PHP track of Exercism. I do not have plenty of time, but could add a few concept exercises per month. Or similar.

Where should I start? Is there a plan for the syllabus? A list of concepts to have? Or something else you prepared to pick up and incorporate?

I saw #324 and maybe I should give that a try first? I'm not sure if the existing concept exercises cover enough topics to lock a good share of the practice exercises, so maybe that's better done later?

neenjaw:
👋🏻 I'd be happy to have your contribution to the PHP syllabus.

I don't have a strict plan for the syllabus, my approach has been to roughly pick them up in order of complexity, determine which concept exercises should precede them then add them to the flow.  I have found that having a strict plan works poorly to develop a syllabus as long as the concept exercises try to stay atomically focussed on a concept, they can be easily reoriented later to fit a better progression as it becomes more clear.

My recommendation would be to pick a concept from `/concepts` that doesn't have a corresponding primary exercise in `/exercises/concept` then go from there.  There are pretty good docs around the general concepts of exercise building [on the website](https://exercism.org/docs/building/product/concept-exercises)

mk-mxp:
I would start with `concepts/type-juggling`, as it nicely fits after the existing Type Declaration. After reading the introduction and the examples in it: How about strict types? These are introduced in Type Declaration (about.md only), but not actively used in there.

As strict types are activated per file, I could make one file without and one file with that turned on. The exercise would be to understand the implicit type conversions happening in the one without strict types and make them explicit in the file with strict types. What do you think?

neenjaw:
I think that while we have the strict type declaration in all of the practice exercises, type juggling is a big part of the language.  I think it is important to develop fluency in PHP to realize how pervasive type juggling can be.

I'm not opposed to a "strict types" concept exercise checkpoint, and every exercise after that one has strict types.

My approach to the current progression is that strong types are an _additional value_ but they are optional. In practice, I would always use types where possible with a strict declaration where possible, but students need to know how to navigate when they don't exist.

Does that perspective help shape your approach?

neenjaw:
No worries, no rush, this is also on volunteer basis for me

mk-mxp:
I've been thinking a lot about where to work next. I came up with this as a possible next goal:

[...]

PR #607, opened 2024-01-30
Overhaul `basic-syntax` and `lasagna`

mk-mxp:
This very first concept and its concept exercise touch a lot of concepts coming later.
These must be introduced as much as necessary to understand the code and the requested solution, but not more.
So we have many short introductions and a bit more depth for the `about.md`. But what is "Basics" or "basic-syntax" then?

The important decision guiding document (according to exercisms docs) is `design.md`, which was a copy from the Ruby track and not adjusted for PHP Basics. I analyzed the exercise (as this must remain mostly the same to not invalidate solutions) and re-wrote `design.md` to inform further decisions. And answered the question "what to teach here" from this perspective.

I then added missing conceptual information to `about.md` for little experienced programmers and condensed that document for `introduction.md` to address medium experienced programmers. I kept the chapter on code organisation in `about.md` and added a chapter on simple debugging. As I think this helps people who are stuck also, I added `var_dump()` to `hints.md`.

I focused the exercise stub on doing the function bodies for the tasks from `instructions.md` by adding method definitions without body. As outlined in `design.md` I think it is not "basic syntax" to define the interface of methods, that is "user-defined functions" and "class basics".

I intentionally chose comments for the function bodies and not throwing exceptions, as that would add another concept to touch.

Finally I added more value to the exercise tests by making their feedback more helpful and guiding.

- [x] The concept has its own concept exercise
- [x] The concepts configuration integrates nicely into the concept map
  - [x] The name and subtext (blurb) read well together
  - [x] ~~The concept is connected to the prerequisite concepts as planned~~ (has none)
- [x] The concept follows exercism guidelines
  - [x] `about.md` for in-depth discussion of the topic
  - [x] `introduction.md` with concise information to address a medium experienced programmer who is new to PHP
  - [x] `introduction.md` links to `about.md` for those who need / want more information
- [x] The concept exercise follows exercism guidelines
  - [x] `design.md` outlines goals, learning objectives, scope, unlocked concepts, prerequisites
  - [x] Re-use introduction from concept
  - [x] `instructions.md` uses steps to drive towards the solution
  - [x] `hints.md` unstucks programmers without handing out the solution, mostly by refering to helpful sources to read
  - [x] `hints.md` links back to `about.md` where reasonable
- [x] The concept exercise uses only concepts already introduced
- [x] The concept exercise adheres to PHP track consensus:
  - [x] Use classes, not functions
  - [x] Names are in unified casing (classes PascalCase, methods camelCase, variables snake_case)
  - [x] Strict types in tests (and optionally in exemplar), but not in students files (the concept to be mastered is still missing)
  - [x] ~~Avoid named arguments to allow students to name parameters as desired~~
- [x] The tests make sense and are helpful to master the exercises
  - [x] Order of tests matches the order of instructions
  - [x] Failing tests give helpful information

Adds to #578

### DIFF
--- a/concepts/basic-syntax/about.md
+++ b/concepts/basic-syntax/about.md
@@ -15,6 +15,9 @@ See [PHP tag docs][php-tag] to learn more about the PHP tag.
 // No closing tag required at the end
 ```
 
+For easier reading, all code examples omit the PHP opening tag `<?php`.
+But we have it in all PHP exercises, so you don't have to add it yourself.
+
 All statements, like assignments or function calls, must end with a `;` for instruction separation.
 Omitting `;` (like in JavaScript) is not allowed and results in syntax errors.
 Learn more about [instruction separation][php-semicolon] from the PHP documentation.
--- a/concepts/basic-syntax/about.md
+++ b/concepts/basic-syntax/about.md
@@ -4,7 +4,7 @@
 
 PHP is an interpreted language, meaning it executes the code you write directly.
 PHP code files are text files that contain a PHP opening tag `<?php` to mark the start of PHP code.
-If the file contains only PHP code, there is no closing tag `?>` to mark the end of PHP code.
+If the file contains only PHP code, it is possible to omit the closing tag `?>` which is used to mark the end of PHP code.
 See [PHP tag docs][php-tag] to learn more about the PHP tag.
 
 ```php
--- a/concepts/basic-syntax/about.md
+++ b/concepts/basic-syntax/about.md
@@ -52,16 +52,17 @@ Literal values and values stored in variables can be passed to functions.
 New variables are created when defining function arguments to hold values passed in.
 
 ```php
-window_width(100); // Pass the literal value 100 to the function
-
-$newWidth = 100;
-window_width($newWidth); // Pass the value stored in $newWidth to the function
-
+// Declare the window_width function
 function window_width($width)
 {
     echo $width; // Access the value passed into the function as $width
     // ...
 }
+
+window_width(100); // Call the function with the literal value 100
+
+$newWidth = 100;
+window_width($newWidth); // Call the function with the value stored in $newWidth
 ```
 
 Functions may return values using the keyword `return`.
--- a/concepts/basic-syntax/about.md
+++ b/concepts/basic-syntax/about.md
@@ -33,8 +33,8 @@ Variable names must start with a dollar `$` sign.
 To dig deeper into variables, see [PHP variable docs][php-variables].
 
 ```php
-$count = 1; // Assigned an integer value of 1
-$count = 2; // Re-assigned a new value to the variable
+$count = 1; // Assign an integer value of 1
+$count = 2; // Re-assign a new value to the variable
 
 $count = false; // You may assign any value type
 
--- a/concepts/basic-syntax/about.md
+++ b/concepts/basic-syntax/about.md
@@ -1,36 +1,141 @@
 # Basics
 
+## General syntax
+
+PHP is an interpreted language, meaning it executes the code you write directly.
+PHP code files are text files that contain a PHP opening tag `<?php` to mark the start of PHP code.
+If the file contains only PHP code, there is no closing tag `?>` to mark the end of PHP code.
+See [PHP tag docs][php-tag] to learn more about the PHP tag.
+
+```php
+<?php
+
+// All PHP code goes here
+
+// No closing tag required at the end
+```
+
+All statements, like assignments or function calls, must end with a `;` for instruction separation.
+Omitting `;` (like in JavaScript) is not allowed and results in syntax errors.
+Learn more about [instruction separation][php-semicolon] from the PHP documentation.
+
+```php
+$message = "Success!"; // Statement correctly ends with `;`
+$message = "I know JavaScript." // PHP Parse error: syntax error, [...]
+```
+
 ## Values and Variables
 
-PHP is a general-purpose dynamically-typed programming language.
+PHP is a dynamically-typed programming language.
 Value types are only checked while the program is running.
 Using the assignment `=` operator, any value type may be assigned to any variable name.
 Variable names must start with a dollar `$` sign.
+To dig deeper into variables, see [PHP variable docs][php-variables].
 
 ```php
-<?php
-$count = 1 // Assigned an integer value of 1
-$count = 2 // Re-assigned a new value to the variable
+$count = 1; // Assigned an integer value of 1
+$count = 2; // Re-assigned a new value to the variable
 
-$count = false // You may assign any value type
+$count = false; // You may assign any value type
 
 // Strings can be created by enclosing the characters
 // within single `'` quotes or double `"` quotes.
-$message = "Success!"
+$message = "Success!";
 ```
 
-## Code organization
+## Functions and Methods
 
-PHP code files are text files that start with `<?php`.
-If the file contains only php code, there is no closing `?>` tag.
-Code, functions, and classes may all be present in a single file, but usually classes are in their own file with a filename which matches the class.
+Literal values and values stored in variables can be passed to functions.
+New variables are created when defining function arguments to hold values passed in.
+
+```php
+window_width(100); // Pass the literal value 100 to the function
+
+$newWidth = 100;
+window_width($newWidth); // Pass the value stored in $newWidth to the function
+
+function window_width($width)
+{
+    echo $width; // Access the value passed into the function as $width
+    // ...
+}
+```
+
+Functions may return values using the keyword `return`.
+Values to return can be literals, stored in variables, or the result of an instruction.
+
+```php
+function window_height()
+{
+    $heightOffset = 10;
+
+    return 100 + $heightOffset;
+}
+```
+
+Functions inside classes and their instances are called methods.
+To call a method, the name is preceeded by the instance and `->`.
+A class instance is created by the `new` operation.
+Methods have access to the special variable `$this`, which refers to the current instance.
 
 ```php
 <?php
-// index.php
+// Calculator.php
 
-$sum = add(1, 2);
+class Calculator {
+    // ...
+
+    public function sub($x, $y)
+    {
+        return $this->add($x, -$y); // Calls the method add() of the current instance
+    }
+
+    public function add($x, $y)
+    {
+        return $x + $y;
+    }
+}
+
+$calculator = new Calculator(); // Creates a new instance of Calculator class
+$calculator->sub(3, 1); // Calls the method sub() of the instance stored in $calculator
+// => 2
 ```
+
+We will dig deeper into functions and classes in later exercises.
+
+## Simple debugging
+
+As a basic form of debugging, printing values and contents of variables helps with understanding what happens.
+PHP can print values using:
+
+- `echo $variable;` to output a string representation of the value
+- `print $variable;`, which is the same as `echo $variable;`
+- `var_dump($variable);` to get more insight into the value, e.g. the type PHP sees
+
+When using `echo` or `print`: Empty strings, `false` and `null` will not show up.
+And classes and instances will make them error out.
+Use `var_dump()` to see these.
+
+In the online editor, the output you produce is shown with the test results.
+When you are using the CLI, the output appears mixed with the progress indicator before test summary.
+
+## Code organization
+
+Files are basic units of code organisation in PHP and many other languages.
+For showing the name of the file to look at in a code block, we use a comment.
+You never have to add such comments to your files:
+
+```php
+<?php
+// NameOfPhpFile.php
+
+[...]
+```
+
+Code, functions, and classes may all be present in a single file.
+But usually these kinds of code units are in separate files.
+
+It is common practice to have one class per file with a filename which matches the class.
 
 ```php
 <?php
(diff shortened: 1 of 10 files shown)


mk-mxp:
I'm not sure about re-arranging the tests to follow [Arrange, Act, Assert](https://wiki.c2.com/?ArrangeActAssert) or [Given, When, Then](https://martinfowler.com/bliki/GivenWhenThen.html) with additionally separated out input and expected values. The arrangement would allow easier implementation of a test generator for practice exercises, students would get used to seeing a test arrangement recommended for unit testing and many implementation details look much nicer when separating the data from the phases.

Current code:

```php
    public function testTotalElapsedTime(): void
    {
        $lasagna = new Lasagna();
        $this->assertEquals(21, $lasagna->totalElapsedTime(layers_to_prep: 4, elapsed_minutes: 13));
    }
```

Arrange-Act-Assert:

```php
    public function testTotalElapsedTime(): void
    {
        $input = [
            'layers_to_prep' => 4,
            'elapsed_minutes' => 13,
        ];
        $expected = 21;

        $subject = new Lasagna();
        $actual = $subject->totalElapsedTime(...$input);

        $this->assertEquals($expected, $actual);
    }
```

Also I would suggest using a common naming scheme for the tests. E.g. I usually use `$expected`, `$actual`, `$subject` (rather than `$subjectUnderTest` or `$sut`) like shown in the example above.

concepts/basic-syntax/about.md
```diff
@@ -1,37 +1,142 @@
 # Basics
 
+## General syntax
+
+PHP is an interpreted language, meaning it executes the code you write directly.
+PHP code files are text files that contain a PHP opening tag `<?php` to mark the start of PHP code.
+If the file contains only PHP code, there is no closing tag `?>` to mark the end of PHP code.
```
homersimpsons, line 7:
```suggestion
If the file contains only PHP code, it is possible to omit the closing tag `?>` which is used to mark the end of PHP code.
```

concepts/basic-syntax/about.md
```diff
@@ -1,37 +1,142 @@
...
+
+// No closing tag required at the end
+```
+
+All statements, like assignments or function calls, must end with a `;` for instruction separation.
+Omitting `;` (like in JavaScript) is not allowed and results in syntax errors.
+Learn more about [instruction separation][php-semicolon] from the PHP documentation.
+
+```php
+$message = "Success!"; // Statement correctly ends with `;`
```
homersimpsons, line 23: I think it is relevant to keep the opening `<?php` tags in this introduction or clearly state that we will omit them for readability.

concepts/basic-syntax/about.md
```diff
@@ -1,37 +1,142 @@
...
+window_width(100); // Pass the literal value 100 to the function
+
+$newWidth = 100;
+window_width($newWidth); // Pass the value stored in $newWidth to the function
+
+function window_width($width)
+{
+    echo $width; // Access the value passed into the function as $width
+    // ...
+}
```
homersimpsons, line 61: It certainly makes more sense to declare the function at the beginning, unless the goal is to showcase that declaration order is not relevant.

```suggestion
// declare the window_width function
function window_width($width)
{
    echo $width; // Access the value passed into the function as $width
    // ...
}

window_width(100); // Call the function with the literal value 100

$newWidth = 100;
window_width($newWidth); // Call the function withthe value stored in $newWidth
```

concepts/basic-syntax/about.md
```diff
@@ -1,37 +1,142 @@
...
+
+    public function add($x, $y)
+    {
+        return $x + $y;
+    }
+}
+
+$calculator = new Calculator(); // Creates a new instance of Calculator class
+$calculator->sub(3, 1); // Calls the method sub() of the instance stored in $calculator
+// => 2
```
homersimpsons, line 101:
```suggestion
$result = $calculator->sub(3, 1); // Calls the method sub() of the instance stored in $calculator
echo $result; // => 2
```

concepts/basic-syntax/about.md
```diff
@@ -1,37 +1,142 @@
...
 
+We will dig deeper into functions and classes in later exercises.
+
+## Simple debugging
+
+As a basic form of debugging, printing values and contents of variables helps with understanding what happens.
+PHP can print values using:
+
+- `echo $variable;` to output a string representation of the value
+- `print $variable;`, which is the same as `echo $variable;`
```
homersimpsons, line 112: I do not find `print` to be common, do we want to show it here ?

concepts/basic-syntax/about.md
```diff
@@ -1,37 +1,142 @@
...
+// NameOfPhpFile.php
 
-$sum = add(1, 2);
+[...]
 ```
 
+Code, functions, and classes may all be present in a single file.
+But usually these kinds of code units are in separate files.
+
+It is common practice to have one class per file with a filename which matches the class.
```
homersimpsons, line 138: Do we want to add a reference to PSR-4 ? It may be too confusing though.

concepts/basic-syntax/introduction.md
```diff
@@ -1,68 +1,79 @@
...
-<?php
-// index.php
+function window_height($height)
+{
+    return $height + 10;
+}
 
-$sum = add(1, 2);
+window_height(100);
+// => 110
```
homersimpsons, line 51:
```suggestion
$result = window_height(100);
echo $result; // => 110
```

exercises/concept/lasagna/.docs/hints.md
```diff
@@ -2,43 +2,44 @@
...
 
 ## 2. Calculate the remaining oven time in minutes
 
-- You need to define a [function][functions] with a single argument.
-- You have to [explicitly return an integer][return] from a function.
-- The function's argument is an [integer][integers].
-- You can use the [mathematical operator for subtraction][operators] to subtract values.
+- You need to [explicitly return][php-return] a value from a function.
+- The function's argument is an [integer][php-integers].
+- You can use the [mathematical operator for subtraction][php-operators] to subtract values.
```
homersimpsons, line 18:
```suggestion
- You can use the [mathematical operator for subtraction][php-operators] `-` to subtract values.
```

exercises/concept/lasagna/.docs/hints.md
```diff
@@ -2,43 +2,44 @@
...
 
 ## 3. Calculate the preparation time in minutes
 
-- You need to define a [function][functions] with a single argument.
-- You have to [implicitly return an integer][return] from a function.
-- The function's argument is an [integer][integers].
-- You can use the [mathematical operator for multiplication][operators] to multiply values.
+- You need to [explicitly return][php-return] a value from a function.
+- The function's argument is an [integer][php-integers].
+- You can use the [mathematical operator for multiplication][php-operators] to multiply values.
```
homersimpsons, line 25:
```suggestion
- You can use the [mathematical operator for multiplication][php-operators] `*` to multiply values.
```

exercises/concept/lasagna/.docs/hints.md
```diff
@@ -2,43 +2,44 @@
...
 
-- You need to define a [function][functions] with two arguments.
-- You have to [implicitly return an integer][return] from a function.
-- The function's argument is an [integer][integers].
-- You can invoke one of the other functions you've defined previously.
-- You can use the [mathematical operator for addition][operators] to add values.
+- You need to [explicitly return][php-return] a value from a function.
+- The function's arguments are [integers][php-integers].
+- You can [invoke one of the other functions][exercism-methods] of the class.
+- You can use the [mathematical operator for addition][php-operators] to add values.
```
homersimpsons, line 32:
```suggestion
- You can use the [mathematical operator for addition][php-operators] `+` to add values.
```

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -2,68 +2,80 @@
...
-<?php
-// index.php
+function window_height($height)
+{
+    return $height + 10;
+}
 
-$sum = add(1, 2);
+window_height(100);
+// => 110
```
homersimpsons, line 53:
```suggestion
$result = window_height(100);
echo $result; // => 110
```

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -2,68 +2,80 @@
...
-<?php
-
-// Single line comment
 
-/*
-  Multi-line comment
-*/
+$calculator = new Calculator(); // Creates a new instance of Calculator class
+$calculator->sub(3, 1); // Calls the method sub() of the instance stored in $calculator
+// => 2
```
homersimpsons, line 77:
```suggestion
$result = $calculator->sub(3, 1); // Calls the method sub() of the instance stored in $calculator
echo $result; // => 2
```

exercises/concept/lasagna/LasagnaTest.php
```diff
@@ -2,49 +2,79 @@
 
 declare(strict_types=1);
 
+/**
+ * We use `assertEquals()` here for its loose type checking. We don't care if
+ * the student returns numeric strings or integers in this exercise.
+ *
+ * - Please use `assertSame()` whenever possible. Note reasons for exceptions.
```
homersimpsons, line 9:
```suggestion
 * - Please use `assertSame()` whenever possible. Add a comment when it is not possible.
```

exercises/concept/lasagna/LasagnaTest.php
```diff
@@ -2,49 +2,79 @@
...
-    /** @task_id 2 */
+    /**
+     * @testdox Returns how many minutes more the lasagna must be in the oven when it is 30 minutes in the oven already
+     * @task_id 2
+     */
     public function testAnotherRemainingCookTime(): void
     {
         $lasagna = new Lasagna();
-        $this->assertEquals(10, $lasagna->remainingCookTime(30));
+        $this->assertEquals(10, $lasagna->remainingCookTime(elapsed_minutes: 30));
```
homersimpsons, line 51: (see named argurment discussion)

exercises/concept/lasagna/LasagnaTest.php
```diff
@@ -2,49 +2,79 @@
...
-    /** @task_id 3 */
+    /**
+     * @testdox Returns how many minutes you spent preparing the lasagna with 7 layers
+     * @task_id 3
+     */
     public function testTotalPreparationTime(): void
     {
         $lasagna = new Lasagna();
-        $this->assertEquals(14, $lasagna->totalPreparationTime(7));
+        $this->assertEquals(14, $lasagna->totalPreparationTime(layers_to_prep: 7));
```
homersimpsons, line 61: (see named argurment discussion)

exercises/concept/lasagna/LasagnaTest.php
```diff
@@ -2,49 +2,79 @@
...
-    /** @task_id 4 */
+    /**
+     * @testdox Returns the total minutes you have worked on the lasagna with 4 layers that is 13 minutes in the oven
+     * @task_id 4
+     */
     public function testTotalElapsedTime(): void
     {
         $lasagna = new Lasagna();
-        $this->assertEquals(21, $lasagna->totalElapsedTime(4, 13));
+        $this->assertEquals(21, $lasagna->totalElapsedTime(layers_to_prep: 4, elapsed_minutes: 13));
```
homersimpsons, line 71: (see named argurment discussion)

homersimpsons (changes requested):
Overall a great improvemen! I opened some threads on things I think we can improve.

Do not hesitate to discuss them.

concepts/basic-syntax/about.md
mk-mxp: I have seen `print` in quite some PHP code, in one company even as must-have in coding styles. As much as I prefer `echo` myself, it seems to _feel better_ for native speakers or users of other languages. This is the only place I mention it, it's not in the `introduction.md` nor in the tracks `debug.md`. So I'd leave it in here for completeness, but not promote it further.

concepts/basic-syntax/about.md
mk-mxp: I am not sure, if adding a reference to a topic like autoloading (PSR-4 is just one convention using files with matching class names) distracts too much from the basics to learn. I would rather take out the whole `Code organization` topic than adding more to it. Are you OK with leaving PSR-4 for the namespace concept?

exercises/concept/lasagna/.docs/hints.md
mk-mxp: I'm not sure about adding the concrete operator to the hints. In the [exercism docs](https://exercism.org/docs/building/tracks/concept-exercises#h-file-docs-hints-md), it says:

> The hints should not spell out the solution, but instead point to a resource describing the solution

To me, adding the concrete operator is _spelling out the solution_. Do you agree?

exercises/concept/lasagna/.docs/introduction.md
mk-mxp: This type of showing the resulting value of the previous operation is used across all PHP concept exercises. In fact, `echo` cannot be used for `false` or `''` or complex values in arrays / objects. So I kept this notation as it is a compromise between code correctness and understandability. And it is [used in other languages as well](https://github.com/exercism/javascript/blob/main/exercises/concept/lasagna/.docs/instructions.md#2-calculate-the-remaining-oven-time-in-minutes).

exercises/concept/lasagna/LasagnaTest.php
mk-mxp: Can you point to a resurce for the "named argurment discussion"? It's not in GitHub issues. Forum? Discord?

exercises/concept/lasagna/.docs/hints.md
homersimpsons: Okay, I'm fine with that. It avoid duplicating the doc and shows the student where to look for ressources.

exercises/concept/lasagna/.docs/introduction.md
homersimpsons: Okay, makes sense then, let's keep it like this and be consistent with the rest.

concepts/basic-syntax/about.md
homersimpsons: > Are you OK with leaving PSR-4 for the namespace concept?

Okay to do it in this concept.

exercises/concept/lasagna/LasagnaTest.php
homersimpsons: > Can you point me to a resource for the "named argument discussion"? It's not in GitHub issues. Forum? Discord?

Woops, looks like I did not comment it in any case.

I'm against using named arguments in the test. In fact, a student may change the arguments' names and then the tests will break.

So to me we should not test that the argument names are correct.

What to you think ?

mk-mxp: Using positional parameters in later concept exercises is my goal, as you can see from the DocBlock above the test class asking for not using them. But I forgot to write down the rational about why I use them here...will add that.

I wanted to use named arguments in the first concept exercises to help students see the connection between the value in the test code and the parameter name of the function. This is because in `about.md`, `introduction.md` and `instructions.md` we introduce the general idea of using the parameter name as a variable in the function body, but not that parameter position is relevant. It is not like this in all languages. And I think that this rather abstract explanation would not stick long enough until required for solving the exercise.

I wanted to introduce these details in `user-defined-functions` concept, when showing named vs. positional parameters using optional parameters. That will be shortly after the basic types in the concept tree. And not use named parameters after that anymore.

homersimpsons: I understand your point, wouldn't this add additional burden if a student changes the parameter name ?

Would it make sense to duplicate the tests to have one with positional argument and the other one with named arguments ?

> This is because in `about.md`, `introduction.md` and `instructions.md` we introduce the general idea of using the parameter name as a variable in the function body, but not that parameter position is relevant.

We are not showing named arguments here. I think if a student reads the test code he may be a bit lost reading the argument names here.

> wanted to introduce these details in `user-defined-functions` concept, when showing named vs. positional parameters using optional parameters. That will be shortly after the basic types in the concept tree. And not use named parameters after that anymore.

Are we in `user-defined-functions` concept here ?

mk-mxp: OK then, named arguments are gone. I still believe in the value of named arguments for understandig which value goes where. But I do not have data to proof my point as well I have no data to proof the opposite.

homersimpsons (approved):

homersimpsons:
Thank you very much for all those improvements !

PR #612, opened 2024-02-06
Adjust tests for Nucleotide Count exercise to match problem-specifications

tomasnorre:
Adjusted tests, to problem specifications [1] and as agreed in forum [2]. Please let me know if changes wanted. I have tested that they work with my locally solved Nucleotide Count Exercise. They are also working locally with the `PHPUNIT_BIN="./bin/phpunit-9.phar" ./bin/test.sh`

1) https://github.com/exercism/problem-specifications/blob/main/exercises/nucleotide-count/canonical-data.json
2) https://forum.exercism.org/t/improvement-of-exercise-nucleotide-count/9651/1

### DIFF
--- a/exercises/practice/nucleotide-count/.meta/config.json
+++ b/exercises/practice/nucleotide-count/.meta/config.json
@@ -5,7 +5,8 @@
   "contributors": [
     "arueckauer",
     "kytrinyx",
-    "petemcfarlane"
+    "petemcfarlane",
+    "tomasnorre"
   ],
   "files": {
     "solution": [
--- a/exercises/practice/nucleotide-count/NucleotideCountTest.php
+++ b/exercises/practice/nucleotide-count/NucleotideCountTest.php
@@ -31,6 +31,9 @@ class NucleotideCountTest extends PHPUnit\Framework\TestCase
         require_once 'NucleotideCount.php';
     }
 
+    /**
+     * uuid: 3e5c30a8-87e2-4845-a815-a49671ade970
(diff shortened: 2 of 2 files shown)


mk-mxp:
There should be a `.meta/tests.toml` file in the exercise directory listing the tests this track has implemented. Is that in sync with your changes? The UUIDs should match with the problem spec and the name with your tests. See [Exercism docs for further info](https://exercism.org/docs/building/configlet/sync#h-tests). If there are differences, you may automatically sync it: `bin/configlet sync --metadata --update --exercise nocleotide-count`. I haven't used that magic myself, yet.

There is also a `.meta/config.json` file in the exercise directory, where you may add yourself to the list of contributors. Use the Exercism / GitHub username here.

One thing I would love to have is adding the UUID as a DocBlock above each test. This is in no way discussed with the current maintainer nor "a rule" within exercism. But in my eyes this would really help linking the pieces together for (yet to be invented) tools like Test Generators and contributors. But we may also keep that for later.

tomasnorre:
> There should be a .meta/tests.toml file in the exercise directory listing the tests this track has implemented. Is that in sync with your changes?

I wasn't aware of that file, will check it and adjust if needed.

> One thing I would love to have is adding the UUID as a DocBlock above each test. 

I can take care of that.

mk-mxp:
@tomasnorre Thanks a lot for contributing!

mk-mxp (approved):

PR #591, opened 2023-12-06
Overhaul tool usage in CI and for contribution

mk-mxp:
- Put all code styling rules into PHPCS config
  - Add XML namespace for completeness
  - Remove non-existent `type="int"`
  - Add directory rules for improved test coverage and required exceptions
  - Fix the new upcoming violations
- PHPCS config renamed to default name
- Moved tool installation to composer
- Moved tool usage to composer
  - Allows running all unit tests now, too
- Overhauled README

This is part of the discussion in [the forum](https://forum.exercism.org/t/php-streamlining-php-codesniffer-usage/8538).

PR #613, opened 2024-02-06
Remove > from shell readme, to ease copy pasting

tomasnorre:
We haven't agreed on this change, so feel free to reject it.

Currently, the GitHub copy function isn't working with the `> ./bin/install.sh` command. 

It will end up in a loop waiting for input, and write an empty `bin/install.sh` file.

By removing the `>` it can be copied directly into a terminal and the commands can be executed as wanted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,19 +10,19 @@ Exercism exercises in PHP
 ### All dependencies
 
 ```shell
-> ./bin/install.sh
+./bin/install.sh
 ```
 
 ### Only tests dependencies
 
 ```shell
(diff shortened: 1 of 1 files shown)


mk-mxp:
This may be obsolete with #591 , which is not yet merged and would change the whole README / contribution tool chain.
Currently it is not decided, which way to go. Thanks for your patience!

homersimpsons:
This was superseded by #591, I close this one. Thank you for the contribution still.

PR #614, opened 2024-02-10
Add debug hints to show up in online editor

mk-mxp:
I discovered that on the JavaScript track lately:

JavaScript online editor

It's easy to add that for PHP, too.

### DIFF
--- a/exercises/shared/.docs/debug.md
+++ b/exercises/shared/.docs/debug.md
@@ -1,0 +1,9 @@
+# Debug
+
+When a test fails, a message is displayed describing what went wrong and for which input.
+
+Printing values and contents of variables can also help with understanding what happens.
+PHP can print values using:
+
+- `echo 'Debug message';` to output a string of your choice
+- `var_dump($variable);` to get more insight into a value, e.g. the type PHP sees


exercises/shared/.docs/debug.md
```diff
@@ -0,0 +1,9 @@
+# Debug
+
+When a test fails, a message is displayed describing what went wrong and for which input.
+
+Printing values and contents of variables can also help with understanding what happens.
+PHP can print values using:
+
+- `echo 'Debug message';` to output a string of your choice
+- `var_dump($variable);` to get more insight into a value, e.g. the type PHP sees
```
homersimpsons, line 9: Does this wok with the curent test-runner ?

Should we also highlight `error_log` and `print_r` ?

homersimpsons (approved):

exercises/shared/.docs/debug.md
mk-mxp: Any output is shown in the online editor, below the test code run. You can test that using the "Hello, world!" exercise in the online editor.

I wanted to keep it small and chose the options that are most fail-safe. `echo` is best for strings, as `false` and such do not show up. `var_dump()` is best for variables, as it shows types and works with arrays and objects. `print_r()` may not show output if one adds `true` as second parameter. `error_log()` is expert level and has lots of possible pitfalls.

PR #615, opened 2024-02-13
Update introduction.md

neilmayhew:
Add a note about the use of `$this`. It's needed to be able to solve the first exercise, and people coming from other languages (eg C++) may not realize it's required rather than optional.

### DIFF
--- a/exercises/concept/lasagna/.docs/introduction.md
+++ b/exercises/concept/lasagna/.docs/introduction.md
@@ -48,6 +48,15 @@ class Calculator {
 }
 ```
 
+To call a method from another method of the same class, you need to use the `$this` variable:
+
+```php
+    function subtract($x, $y)
+    {
+        return $this->add($x, -$y);
+    }
+```
+
 ### Naming conventions
 
 Classnames should all be `PascalCase`.


mk-mxp:
Thanks for pointing that out, @neilmayhew ! This MR does conflict with the rewriting of the whole concept in #607. Would you please take a look at [the re-written lines](https://github.com/exercism/php/pull/607/files#diff-7da82573943da07b2bbdd45d31909e270ca73554c78014a56e112ccc7017a441R56) and tell me, if that clarifies the issue enough or not?

neilmayhew:
> This MR does conflict with the rewriting of the whole concept in #607. Would you please take a look at [the re-written lines](https://github.com/exercism/php/pull/607/files#diff-7da82573943da07b2bbdd45d31909e270ca73554c78014a56e112ccc7017a441R56) and tell me, if that clarifies the issue enough or not?

Yes, that looks great!

My only concern is that the material is duplicated in several places, but I imagine that's unavoidable in the existing Exercism framework.

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -48,6 +48,15 @@ class Calculator {
 }
 ```
 
+To call a method from another method of the same class, you need to use the `$this` variable:
+
+```php
+    function subtract($x, $y)
+    {
+        return $this->add($x, -$y);
+    }
```
homersimpsons, line 57: What is the reason to indent this ?

Should we show the entire `class Calculator { ... }` ?

homersimpsons (approved):

mk-mxp:
> My only concern is that the material is duplicated in several places, but I imagine that's unavoidable in the existing Exercism framework.

Yes, you are right. Exercism has automation to fill `introduction.md` in an exercise from another document in the `concept` directory.

mk-mxp:
@homersimpsons Are you OK with giving @neilmayhew a tiny amount of reputation by merging this as is and then merging #607 with the complete rewrite?

exercises/concept/lasagna/.docs/introduction.md
neilmayhew: The indentation was indeed to show that it follows on from the preceding example. It could be put into the preceding, and then the note could refer back to it.

However, I assume this is moot now that #607 is going to rewrite everything.

neilmayhew:
I see that you assigned this to me, but I'm not sure what I'm supposed to do, since I don't have merge rights to the repo. Did you mean to assign it to @homersimpsons instead?

homersimpsons:
Okay, I will merge this one then review #607

PR #619, opened 2024-02-18
Auto format all files with configlet

mk-mxp:
Now it should not do that when creating a new exercise.

### DIFF
--- a/config.json
+++ b/config.json
@@ -19,98 +19,120 @@
     "average_run_time": 2
   },
   "files": {
-    "solution": ["%{pascal_slug}.php"],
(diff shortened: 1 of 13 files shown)


homersimpsons:
Of course this is not really reviewable.

@mk-mxp Can we add any CI job to ensure this formating is kept in next Pull Requests ?

mk-mxp:
> @mk-mxp Can we add any CI job to ensure this formating is kept in next Pull Requests ?

I think we need to coordinate this with exercism officials. The configlet workflow is owned by them, and changes are made by a bot.

PR #621, opened 2024-02-20
changing function names to match the instructions

kuprelweronika:
the previous functions name was wrong, in instructions will be shorter version


mk-mxp:
Hello and welcome, Weronika!

When we change the function names as you suggest, all the previous solutions to the exercise are invalid. So we must keep the names in that file.

You could adjust the `.docs/instructions.md` to match the PHP stub. In the code example the name matches the PHP file, only in the explaining text it differs. Do you want to do that?

kuprelweronika:
I came back this names in php code and change names in .docs/instructions.md

PR #622, opened 2024-02-20
Add configlet fmt to CI

mk-mxp:
See: https://forum.exercism.org/t/add-configlet-fmt-to-ci/9869/2


mk-mxp:
@homersimpsons FYI: [The TOML is needed, too](http://forum.exercism.org/t/add-configlet-fmt-to-ci/9869/4). So I added that and merged.

PR #623, opened 2024-02-21
Add Zebra Puzzle Exercise

tomasnorre:
After the suggestion in https://forum.exercism.org/t/implement-some-missing-php-exercises-for-48in24/9865/5 I toke the liberty to create the Zebra Puzzle Exercise.

As this is my first exercise contribution, please let me know what can be improved.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1155,6 +1155,14 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 6
+      },
+      {
+        "slug": "zebra-puzzle",
+        "name": "Zebra Puzzle",
+        "uuid": "8a87919b-3bd0-4fc8-9774-daecd148de42",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 6
       }
     ]
   },
--- a/config.json
+++ b/config.json
@@ -1162,7 +1162,7 @@
         "uuid": "8a87919b-3bd0-4fc8-9774-daecd148de42",
         "practices": [],
         "prerequisites": [],
-        "difficulty": 6
+        "difficulty": 7
       }
     ]
   },
--- a/exercises/practice/zebra-puzzle/.meta/example.php
+++ b/exercises/practice/zebra-puzzle/.meta/example.php
@@ -1,0 +1,223 @@
+<?php
+
+/*
+ * By adding type hints and enabling strict type checking, code can become
+ * easier to read, self-documenting and reduce the number of potential bugs.
+ * By default, type declarations are non-strict, which means they will attempt
+ * to change the original type to match the type specified by the
+ * type-declaration.
+ *
+ * In other words, if you pass a string to a function requiring a float,
+ * it will attempt to convert the string value to a float.
+ *
+ * To enable strict mode, a single declare directive must be placed at the top
+ * of the file.
+ * This means that the strictness of typing is configured on a per-file basis.
+ * This directive not only affects the type declarations of parameters, but also
+ * a function's return type.
+ *
+ * For more info review the Concept on strict type checking in the PHP track
+ * <link>.
+ *
+ * To disable strict typing, comment out the directive below.
+ */
+
+declare(strict_types=1);
+
+class ZebraPuzzle
+{
+    private $waterDrinker = '';
+    private $zebraOwner = '';
+
+    private const FIRST = 1;
+    private const MIDDLE = 3;
+
+    private $red = 0;
+    private $green = 0;
+    private $ivory = 0;
+    private $yellow = 0;
+    private $blue = 0;
+
+    private $englishman = 0;
+    private $spaniard = 0;
+    private $ukrainian = 0;
+    private $japanese = 0;
+    private $norwegian = 0;
+
+    private $coffee = 0;
+    private $tea = 0;
+    private $milk = 0;
+    private $orangeJuice = 0;
+    private $water = 0;
+
+    private $oldGold = 0;
+    private $kools = 0;
+    private $chesterfields = 0;
+    private $luckyStrike = 0;
+    private $parliaments = 0;
+
+    private $dog = 0;
+    private $snails = 0;
+    private $fox = 0;
+    private $horse = 0;
+    private $zebra = 0;
+
+    private $nationalityNames = [];
+
+    private $possiblePermutations;
+
+    public function __construct()
+    {
+        $this->possiblePermutations = $this->permuteValues([1, 2, 3, 4, 5]);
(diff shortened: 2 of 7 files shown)


config.json
```diff
@@ -1155,6 +1155,14 @@
...
         "prerequisites": [],
         "difficulty": 6
+      },
+      {
+        "slug": "zebra-puzzle",
+        "name": "Zebra Puzzle",
+        "uuid": "8a87919b-3bd0-4fc8-9774-daecd148de42",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 6
```
ErikSchierboom, line 1165: Maybe increase the difficulty a bit?

ErikSchierboom (commented):
Looks good from an Exercism point of view. I'll let someone from @exercism/php review the code.

config.json
ErikSchierboom: Maybe 7? To me it is slightly more difficult than e.g. `list-ops`, but that could be me 🤷

tomasnorre: I'll add a 7 till someone else have a different suggestion. 6 was the highest till now.. Checked only for 7, 8, 9 and 10. So that's why I went for 6. But can go for 7. :) I'm fine with that. I would even consider having it "hard" instead of just "medium", but I don't know where this label comes from.

exercises/practice/zebra-puzzle/ZebraPuzzleTest.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->zebraPuzzle = new ZebraPuzzle();
+    }
+
+    /**
+     * uuid: 16efb4e4-8ad7-4d5e-ba96-e5537b66fd42
+     */
+    public function testResidentWhoDrinksWater(): void
+    {
+        $this->assertEquals('Norwegian', $this->zebraPuzzle->waterDrinker());
+    }
```
neenjaw, line 47: I really like this type of problem, but I think the interface could be improved because the test interface encourages the return of a single value rather than a solution set that we can interrogate answers from.

I would be interested if maybe we could provide a defined interface for the solution to meet so that we could do something like this:

```php
    // This is me imagineering:
    public function testResidentWhoDrinksWater(): void
    {
        $solution = $this->zebraPuzzle->solve();

        $this->assertEquals('Norwegian', $solution->waterDrinker());
    }
```

I'd even be interested to see if there is a way we could make the problem configurable to be able to have further logic constraint problems if we wanted to extend this problem.

---

Both of those desires are outside of the norm for this problem for other tracks, so it's not required but I think this would differentiate and potentially provide a path to improve this problem.

The solution as it appears here is reasonable given how the problem spec is worded and how other tracks have generally implemented it.

neenjaw (approved):
Have a comment in the test pertaining to how maybe we could advance/differentiate this problem from other tracks, but the problem definition and implementation look reasonable compared to how this problem is generally implemented.

exercises/practice/zebra-puzzle/ZebraPuzzleTest.php
tomasnorre: I think I see what you mean and would be OK with that, but my main question is. Is this for now or for later? If for now we need to agree on which approach we are going to take and is this something we should do for all tracks or only this track?

neenjaw: The scope is for this track, and I believe there is enough leeway that language tracks can differentiate in their interpretation of an exercise. I also don't believe that all tracks need to be identical to teach language fluency.  

As for your question "for now or later", you already have my approval so you don't have to act on any of this.

I just wanted to have a discussion about what my vision of this exercise could be for this track having solved this problem in a few languages.

mk-mxp: I prefer keeping it as is. This is according to the problem spec, as I would read them. And it is the way we translate `"property": "...",` into tests everywhere. The difference against other exercises is more that there is no input to vary for ensuring a real solution.

Regarding the preference of `$solution = $this->zebraPuzzle->solve();` against `$this->zebraPuzzle->waterDrinker();`: I do not want to discuss that here.

exercises/practice/zebra-puzzle/.meta/example.php
```diff
@@ -0,0 +1,223 @@
...
+    {
+        $result = [];
+
+        $length = count($array);
+
+        if ($length === 0) {
+            return [[]];
+        }
+
+        foreach ($array as $index => $Value) {
```
mk-mxp, line 95: Note to self: How could this pass the CI?

```suggestion
        foreach ($array as $index => $value) {
```

mk-mxp (changes requested):
Please update the `main` branch of your repository. We made improvements to the PHP CodeSniffer rules, that should not add the strict types comments to `example.php` and the tests.

exercises/practice/zebra-puzzle/ZebraPuzzleTest.php
neenjaw: > I prefer keeping it as is. This is according to the problem spec, as I would read them. And it is the way we translate "property": "...", into tests everywhere. The difference against other exercises is more that there is no input to vary for ensuring a real solution.

The problem specification for this problem is very old and doesn't describe the parameters of this problem very well. I think if we recognize that it doesn't fit well here, we have the agency to improve it for our track's benefit.

config.json
mk-mxp: I see it as very hard if one is not used to such a kind of puzzle. It is not that hard in respect to PHP knowledge. Anyways I see more an 8 here as anywhere else, as it includes creating a solving strategy for logic puzzles.

exercises/practice/zebra-puzzle/.meta/example.php
tomasnorre: I have updated the value variable.

mk-mxp:
> Can I do anything to check and ensure the configlet step doesn't fail locally?

You can get a local `configlet` using `bin/fetch-configlet`. The run `bin/configlet fmt --update --yes` to apply changes to the files.



> > Please update the `main` branch of your repository. We made improvements to the PHP CodeSniffer rules, that should not add the strict types comments to `example.php` and the tests.
> 
> Should I remove it from the `example.php`? Or doesn't it do any harm?

It does no harm. But it should be inserted automatically only into student-facing files.

mk-mxp (approved):
Thanks a lot everyone for contributing! The track hasn't seen that much diverse activity since long.

PR #624, opened 2024-02-21
Add Circular Buffer Exercise

tomasnorre:
We didn't agree on it, but I found a little time to implement this exercise too. 
Hope that's OK? 

I have linked in on the forum too. 

https://forum.exercism.org/t/implement-some-missing-php-exercises-for-48in24/9865

### DIFF
--- a/exercises/practice/circular-buffer/.meta/example.php
+++ b/exercises/practice/circular-buffer/.meta/example.php
@@ -28,6 +28,10 @@ class BufferFullError extends Exception
 {
 }
 
+class BufferEmptyError extends Exception
+{
+}
+
 class CircularBuffer
 {
     private int $capacity;
@@ -43,10 +47,13 @@ class CircularBuffer
         $this->writePosition = 0;
     }
 
+    /**
+     * @throws BufferEmptyError
+     */
     public function read()
     {
         if ($this->isEmpty()) {
-            return null;
+            throw new BufferEmptyError();
         }
         $value = $this->buffer[$this->readPosition];
         $this->buffer[$this->readPosition] = null;
@@ -67,6 +74,7 @@ class CircularBuffer
     }
 
     /**
+     * @throws BufferEmptyError
      * @throws BufferFullError
      */
     public function forceWrite($item): void
--- a/exercises/practice/circular-buffer/.meta/example.php
+++ b/exercises/practice/circular-buffer/.meta/example.php
@@ -28,10 +28,6 @@ class BufferFullError extends Exception
 {
 }
 
-class BufferEmptyError extends Exception
-{
-}
-
 class CircularBuffer
 {
     private int $capacity;
@@ -47,13 +43,10 @@ class CircularBuffer
         $this->writePosition = 0;
     }
 
-    /**
-     * @throws BufferEmptyError
-     */
     public function read()
     {
         if ($this->isEmpty()) {
-            throw new BufferEmptyError();
+            return null;
         }
         $value = $this->buffer[$this->readPosition];
         $this->buffer[$this->readPosition] = null;
@@ -74,7 +67,6 @@ class CircularBuffer
     }
 
     /**
-     * @throws BufferEmptyError
      * @throws BufferFullError
      */
     public function forceWrite($item): void
--- a/exercises/practice/circular-buffer/.meta/example.php
+++ b/exercises/practice/circular-buffer/.meta/example.php
@@ -1,0 +1,104 @@
+<?php
+
+/*
+ * By adding type hints and enabling strict type checking, code can become
+ * easier to read, self-documenting and reduce the number of potential bugs.
+ * By default, type declarations are non-strict, which means they will attempt
+ * to change the original type to match the type specified by the
+ * type-declaration.
+ *
+ * In other words, if you pass a string to a function requiring a float,
+ * it will attempt to convert the string value to a float.
+ *
+ * To enable strict mode, a single declare directive must be placed at the top
+ * of the file.
+ * This means that the strictness of typing is configured on a per-file basis.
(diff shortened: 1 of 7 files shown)


exercises/practice/circular-buffer/.meta/example.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * @throws BufferEmptyError
+     */
+    public function read()
+    {
+        if ($this->isEmpty()) {
+            throw new BufferEmptyError();
+        }
```
neenjaw, line 57: Suggestion: I think the buffer being potentially empty is a normal case rather than an exceptional case, can this return null instead?

neenjaw (approved):
I'm approving optimistically, but I think it would be an improvement to remove the exception from the read method.

exercises/practice/circular-buffer/.meta/example.php
tomasnorre: It can return whatever we want ;) So we just need to agree on it. I think you are right. I'll add the null return instead unless someone disagrees.

tomasnorre: According to this it should fail. 
https://github.com/exercism/problem-specifications/blob/main/exercises/circular-buffer/canonical-data.json#L28 So it's according to specs.

neenjaw: You're the author of this exercise, would encourage you to accept agency.  I see my role to support your contribution, not gate keep unless my suggestions are accepted.

You've done a good job with these.  I appreciate your contributions.

neenjaw: > According to this it should fail. https://github.com/exercism/problem-specifications/blob/main/exercises/circular-buffer/canonical-data.json#L28 So it's according to specs.

The specs aren't perfect, and so if they advise a pattern which is against "good php implementation and design" I would encourage us to adapt it where we should.

tomasnorre: FYI: I don't see it as criticism or anything alike. I see it as a dialogue for improvement, so that we can give the users the best possible platform for learning. 

And I think you are right, null is more correct than an exception. So I'll adjust it.

mk-mxp: I'm sorry that we are again getting into a discussion that causes @tomasnorre more work.

I'm for using an exception :wink:  I don't see `null` as anything good, not in PHP nor anywhere else. Throwing an exception is good practice, when something should fail for someone doing something unwanted / unfulfillable.

But in my eyes it is more about a "business decision" here. And please use the same kind of failure indication for both, reading an empty and writing to a full buffer.

Any solution is acceptable and according to spec, as long as you are not silently returning a possibly valid value. `null` is as far as I can see a valid value in that implementation. No type specification. No other restriction. So this speaks against `null` as reading failure. When adding `int` as type declaration (as the test cases are all ints), `null` is possible as a failure indicator. And then writing needs the same kind of failure indication (returning a success / failure value, not throwing).

So, choose your path. Both ways are OK, but choose one.

exercises/practice/circular-buffer/CircularBufferTest.php
```diff
@@ -0,0 +1,194 @@
...
+     */
+    public function testReadingEmptyBufferShouldFail(): void
+    {
+        $buffer = new CircularBuffer(1);
+        $this->assertNull($buffer->read());
+    }
+
+    /**
+     * uuid: 2e6db04a-58a1-425d-ade8-ac30b5f318f3
+     * @throws BufferFullError
```
mk-mxp, line 22: `@throws` is not documented for PHPUnit V9.6 in the PHPUnit docs. And I don't know it for using exception suppression. Do you have a link to a source, if it will be supported in PHPUnit 10+?

mk-mxp (changes requested):
Thanks for contributing! I love your flow :smile:

exercises/practice/circular-buffer/.meta/example.php
neenjaw: I'm not sorry for a comment that produces a valuable discussion. Whether we want to encode exception-directed control flow is a reasonable thing to discuss as it pertains specifically to the php implementation.

I don't agree that null doesn't have a place in php though, there is no standard Optional pattern (which supersedes the null-checking patterns in many modern languages -- Rust, Haskell, etc) so there is reasonably a pattern to expressing "nothingness" is a non-exceptional way.

Having worked in large php code bases where everything is considered an exception is cumbersome. 

I agree though (as indicated by my previous approval of the PR), that this is a business concern which @tomasnorre has agency to decide on.

exercises/practice/circular-buffer/CircularBufferTest.php
tomasnorre: It's my IDE suggestion it, so I would have to look it up.

I know that PHPUnit 10 has switched to attributes annotation, and I think they don't have a throw annotation any more. But would have to check up on that.

mk-mxp: Please don't waste time. PHPUnit never had `@throws`. Take those out, they are useless.

exercises/practice/circular-buffer/.meta/example.php
tomasnorre: I have "made" a decision, and here is my argument. 

Under the presumption that both `read()` and `write()` should have the same kind of "error"-handling, I will opt for `Exceptions` it doesn't make any sense to me, to return null, as a value if `write()` cannot be done, due to buffer full, and exception here is more correct in my eyes.

mk-mxp (approved):
Great piece of work! Thanks a lot for contributing.

PR #625, opened 2024-02-21
Add Spiral Matrix Exercise

tomasnorre:
I was just in the flow. Hope that's OK? I'll post a link in the forum too. 

https://forum.exercism.org/t/implement-some-missing-php-exercises-for-48in24/9865

### DIFF
--- a/exercises/practice/spiral-matrix/.meta/example.php
+++ b/exercises/practice/spiral-matrix/.meta/example.php
@@ -26,7 +26,7 @@ declare(strict_types=1);
 
 class SpiralMatrix
 {
-    public function draw($n): array
+    public function draw(int $n): array
     {
         // Initialize the matrix with zeros
         $matrix = array_fill(0, $n, array_fill(0, $n, 0));
--- a/exercises/practice/spiral-matrix/.meta/example.php
+++ b/exercises/practice/spiral-matrix/.meta/example.php
@@ -1,0 +1,80 @@
+<?php
+
+/*
+ * By adding type hints and enabling strict type checking, code can become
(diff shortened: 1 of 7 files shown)


neenjaw (approved):
LGTM

exercises/practice/spiral-matrix/.meta/example.php
```diff
@@ -0,0 +1,80 @@
...
+ * <link>.
+ *
+ * To disable strict typing, comment out the directive below.
+ */
+
+declare(strict_types=1);
+
+class SpiralMatrix
+{
+    public function draw($n): array
```
mk-mxp, line 29:
```suggestion
    public function draw(int $n): array
```

exercises/practice/spiral-matrix/SpiralMatrix.php
```diff
@@ -0,0 +1,33 @@
...
+ * <link>.
+ *
+ * To disable strict typing, comment out the directive below.
+ */
+
+declare(strict_types=1);
+
+class SpiralMatrix
+{
+    public function draw($n): array
```
mk-mxp, line 29:
```suggestion
    public function draw(int $n): array
```

mk-mxp (approved):
These are both optional. Resolve the threads, if you want me to merge it as is.

exercises/practice/spiral-matrix/.meta/example.php
tomasnorre: I have added the type for both. Like it better that way, but completely overlooked it.

PR #634, opened 2024-02-22
Add Secret Handshake Exercise

tomasnorre:
We didn't agree on this, either. But here it is. 

http://forum.exercism.org/t/implement-some-missing-php-exercises-for-48in24/9865/11

### DIFF
--- a/exercises/practice/secret-handshake/SecretHandshakeTest.php
+++ b/exercises/practice/secret-handshake/SecretHandshakeTest.php
@@ -1,0 +1,112 @@
+<?php
+
+declare(strict_types=1);
+
+class SecretHandshakeTest extends PHPUnit\Framework\TestCase
+{
+    private SecretHandshake $secretHandshake;
+
+    public static function setUpBeforeClass(): void
+    {
+        require_once 'SecretHandshake.php';
+    }
+
+    public function setUp(): void
+    {
+        $this->secretHandshake = new SecretHandshake();
+    }
+
+    /**
+     * uuid: b8496fbd-6778-468c-8054-648d03c4bb23
+     */
+    public function testWinkForOne(): void
+    {
+        $this->assertEquals(['wink'], $this->secretHandshake->commands(1));
+    }
+
+    /**
+     * uuid: 83ec6c58-81a9-4fd1-bfaf-0160514fc0e3
+     */
+    public function testDoubleBlinkForTen(): void
+    {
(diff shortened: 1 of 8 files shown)


exercises/practice/secret-handshake/SecretHandshakeTest.php
```diff
@@ -0,0 +1,112 @@
...
+        $this->assertEquals(['wink'], $this->secretHandshake->commands(1));
+    }
+
+    /**
+     * uuid: 83ec6c58-81a9-4fd1-bfaf-0160514fc0e3
+     */
+    public function testDoubleBlinkForTen(): void
+    {
+        $this->assertEquals(['double blink'], $this->secretHandshake->commands(2));
+    }
```
mk-mxp, line 33: I think we could use binary notation here. If students look into the tests to see the actual values passed in, seeing the binary notation helps understanding the task. They will be busy with figuring out the bits in the values when they `echo` or `var_dump()`.

Or maybe mix binary and decimal notation to add some head scratching.

```suggestion
    public function testDoubleBlinkForTen(): void
    {
        $this->assertEquals(['double blink'], $this->secretHandshake->commands(0b0_0010));
    }
```

mk-mxp (approved):
It looks very nice already. If you do not want to add some binary notation: I'm happy to merge it as, too.

exercises/practice/secret-handshake/SecretHandshakeTest.php
tomasnorre: If we change this, the `commands` signature needs to be updated to, currently and `int` is expected. But in general I agree with you, but according to specs it's always `int` values parsed, so the question is should the specs be updated too? If not, I'll keep it as is.

mk-mxp: Well, `0b0_0000` **is** an int. It is PHPs binary integer notation since at least 2022.

tomasnorre: Oh.! I wasn't aware of that, then I'm OK with your change. I'll commit suggestion.

tomasnorre:
> @tomasnorre Maybe a bit more binary notation here and there? Randomly choose test input, so we get a nice mixture of both notations.

I can do that. Hold on ;)

mk-mxp (approved):
Yeah! Very nice.

PR #635, opened 2024-02-22
Add Rotational Cipher Exercise

tomasnorre:
We didn't agree on this, either. But here it is.

http://forum.exercism.org/t/implement-some-missing-php-exercises-for-48in24/9865/11

### DIFF
--- a/exercises/practice/rotational-cipher/.docs/instructions.md
+++ b/exercises/practice/rotational-cipher/.docs/instructions.md
@@ -1,0 +1,29 @@
+# Instructions
+
+Create an implementation of the rotational cipher, also sometimes called the Caesar cipher.
+
+The Caesar cipher is a simple shift cipher that relies on transposing all the letters in the alphabet using an integer key between `0` and `26`.
+Using a key of `0` or `26` will always yield the same output due to modular arithmetic.
+The letter is shifted for as many values as the value of the key.
+
+The general notation for rotational ciphers is `ROT + <key>`.
+The most commonly used rotational cipher is `ROT13`.
+
+A `ROT13` on the Latin alphabet would be as follows:
+
+```text
+Plain:  abcdefghijklmnopqrstuvwxyz
+Cipher: nopqrstuvwxyzabcdefghijklm
+```
+
+It is stronger than the Atbash cipher because it has 27 possible keys, and 25 usable keys.
+
+Ciphertext is written out in the same formatting as the input including spaces and punctuation.
+
+## Examples
+
+- ROT5 `omg` gives `trl`
+- ROT0 `c` gives `c`
+- ROT26 `Cool` gives `Cool`
+- ROT13 `The quick brown fox jumps over the lazy dog.` gives `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.`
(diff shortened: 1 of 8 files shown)


exercises/practice/rotational-cipher/.docs/instructions.md
```diff
@@ -0,0 +1,29 @@
...
+
+Ciphertext is written out in the same formatting as the input including spaces and punctuation.
+
+## Examples
+
+- ROT5 `omg` gives `trl`
+- ROT0 `c` gives `c`
+- ROT26 `Cool` gives `Cool`
+- ROT13 `The quick brown fox jumps over the lazy dog.` gives `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.`
+- ROT13 `Gur dhvpx oebja sbk whzcf bire gur ynml qbt.` gives `The quick brown fox jumps over the lazy dog.`
```
mk-mxp, line 29: OK, ROT13 is in the tests. Maybe we should add a statement against using `str_rot13()` or other PHP encoding functions. Do not append it to this file, as this breaks the sync between problem specification and the track exercise.

Add a new file `.docs/instructions.append.md` with something like this:

```markdown
~~~~exercism/note
Do not use PHP functions like `str_rot13()` to solve this exercise! Only use basic string and character manipulation functions.
~~~~

mk-mxp (changes requested):
Thanks a lot! And such a high quality of your contribution. Your code is nice to read.

exercises/practice/rotational-cipher/.docs/instructions.md
neenjaw: I disagree with strictly preventing some solutions that utilize built in parts of the language. 

I have seen this approach on other tracks and it tends to evolve into a race where maintainers have to then built specific tooling to catch "cheaters" or disable the standard library which is a waste of our limited time. These students only cheat themselves and I think we can encourage the behaviour we want to see positively. 

I think this could be worded:

> PHP has built-in functions to perform string encoding. For an additional challenge, try to solve this problem without using `str_rot13()`.

I think it would also be fair to add Ana additional warning to not use these functions if you need secure encoding.

tomasnorre:
> Thanks a lot for contributing! You must tell me, which AI you use to produce that good code so fast...

ChatGPT, it was that obvious. :) I'm taking existing exercises from other languages and convert the tests and example, and adjust to my liking. It's faster this way.

tomasnorre:
> By default all prs are scored as a medium, I think mkmxp scored some of these as a large? Without tags they don't get attributed to a specific sub-category.

No Problem, thought it has something to do with the tagging, and you confirmed that.

mk-mxp:
@tomasnorre Yes, all my bad. Was too much in hurry this morning. Will take more time and be more cautious.

neenjaw:
Tags can be added retroactively and it will update.

tomasnorre:
> @tomasnorre Yes, all my bad. Was too much in hurry this morning. Will take more time and be more cautious.

Not a huge deal, but must admit the gamification motivates me too :)

mk-mxp:
Thanks for the info, @neenjaw !

PR #639, opened 2024-02-25
Add Eliuds Eggs exercise

homersimpsons:
- [x] Should I add instruction about binary operators ? => yes (https://github.com/exercism/php/pull/639#issuecomment-1963004607)

### DIFF
--- a/exercises/practice/eliuds-eggs/.meta/config.json
+++ b/exercises/practice/eliuds-eggs/.meta/config.json
@@ -1,0 +1,19 @@
+{
+  "authors": [
+    "homersimpsons"
+  ],
+  "files": {
+    "solution": [
+      "EliudsEggs.php"
+    ],
+    "test": [
+      "EliudsEggsTest.php"
+    ],
+    "example": [
+      ".meta/example.php"
+    ]
+  },
+  "blurb": "Help Eliud count the number of eggs in her chicken coop by counting the number of 1 bits in a binary representation.",
+  "source": "Christian Willner, Eric Willigers",
+  "source_url": "https://forum.exercism.org/t/new-exercise-suggestion-pop-count/7632/5"
+}
--- a/exercises/practice/eliuds-eggs/.meta/example.php
+++ b/exercises/practice/eliuds-eggs/.meta/example.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+class EliudsEggs
+{
+    public function eggCount(int $displayValue): int
+    {
+        if ($displayValue === 0) {
+            return 0;
+        }
+
+        return ($displayValue & 1) + $this->eggCount($displayValue >> 1);
+    }
+}
--- a/exercises/practice/eliuds-eggs/EliudsEggsTest.php
+++ b/exercises/practice/eliuds-eggs/EliudsEggsTest.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * By adding type hints and enabling strict type checking, code can become
+ * easier to read, self-documenting and reduce the number of potential bugs.
+ * By default, type declarations are non-strict, which means they will attempt
+ * to change the original type to match the type specified by the
+ * type-declaration.
+ *
+ * In other words, if you pass a string to a function requiring a float,
+ * it will attempt to convert the string value to a float.
+ *
+ * To enable strict mode, a single declare directive must be placed at the top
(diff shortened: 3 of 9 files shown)


exercises/practice/eliuds-eggs/.meta/config.json
```diff
@@ -0,0 +1,17 @@
+{
+  "authors": [],
```
tomasnorre, line 2: You forgot to add you to authors.

exercises/practice/eliuds-eggs/.meta/example.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+/*
```
tomasnorre, line 3: You can remove the comment from files not facing the users.

tomasnorre (approved):
LGTM, with few adjustments.

tomasnorre:
> * [ ]  Should I add instruction about binary operators ?

I personally think that would be helpful, I would appreciate information on that, to easier solve the exercise.

exercises/practice/eliuds-eggs/EliudsEggsTest.php
```diff
@@ -0,0 +1,69 @@
...
+
+class EliudsEggsTest extends PHPUnit\Framework\TestCase
+{
+    public static function setUpBeforeClass(): void
+    {
+        require_once 'EliudsEggs.php';
+    }
+
+    /**
+     * @testdox 0 eggs
```
tomasnorre, line 35: I don't know if the UUID should be in the docblock? Like the exercises I did lately?

homersimpsons: > I don't know if the UUID should be in the docblock? Like the exercises I did lately?

I think we can add it for completness, but it is not used by any tool for now. It may be great to have an `@exercism-uuid` annotation to ensure completness of the tests (we could even extract some informations from this)

> Is @testdox needed, when no dataprovider?

I do not know

> Will phpunit --testdox not output test 0 egges anyway?

Yes it will, I just put it for completness

exercises/practice/eliuds-eggs/.meta/config.json
homersimpsons: Added, thank you

exercises/practice/eliuds-eggs/.meta/example.php
homersimpsons: In fact, I thought this file was linted too

homersimpsons:
> > * [ ]   Should I add instruction about binary operators ?
> 
> I personally think that would be helpful, I would appreciate information on that, to easier solve the exercise.

I did add some instructions in: https://github.com/exercism/php/pull/639/files#diff-1b8add07daeefcbb201624c607300ed14d8a926741b6a6d9562b7faa1c9b291f

Do you think this is enough?

exercises/practice/eliuds-eggs/EliudsEggsTest.php
mk-mxp: > I think we can add it for completness, but it is not used by any tool for now

I'd like to have it in there, as it makes searching tests.toml <-> test.php must more reliable than the title / description. But there is no rule to do so.



> @testdox needed

They are useful. While PHPUnit does a good job on formating the test method name to a readable string for the command line, the JUnit reporter does not. But it outputs the testdox message and that's what is used for the online editor. So many students would see `test0Eggs` only, and for long test method names that does not wrap and gets truncated. So, yes, `@testdox` are some kind of needed. But they are not enforced or required by now.

mk-mxp (approved):
Nice recursive solution to the problem. Add the UUIDs if you want to, but we can merge it as is.

mk-mxp:
> > > * [ ]    Should I add instruction about binary operators ?
> 
> I did add some instructions in: https://github.com/exercism/php/pull/639/files#diff-1b8add07daeefcbb201624c607300ed14d8a926741b6a6d9562b7faa1c9b291f
> 
> Do you think this is enough?

I think that is fine. One does not need much knowledge about binary operators, the source will be enough.

exercises/practice/eliuds-eggs/EliudsEggsTest.php
homersimpsons: Okay, I did add the `uuid` then, I may work on a tool to check them later.

Issue #630, opened 2024-02-22
Tracking: Missing exercises for 48in24


mk-mxp:
These are also tracked in #629 if we cannot add them in time. The week numbers are counted from the start of 48in24, which was on Tuesday of calendar week 3. See [Exercism Implementation Status](https://exercism.org/challenges/48in24/implementation_status) also.

- [ ] ~~W9: parallel-letter-frequency~~ (PHP has no parallel execution)
- [x] W27: secret-handshake #634
- [x] W29: kindergarten-garden #632
- [x] W31: knapsack #633
- [x] W35: rotational-cipher #635
- [x] W39: two-bucket #636
- [x] W42: food-chain #637 
- [x] W43: eliuds-eggs #639

tomasnorre:
As I cannot edit I'll link to the PR here https://github.com/exercism/php/pull/634

homersimpsons:
Every exercise is down now !

Congratulations to all, specifically to @tomasnorre who did much of it !

mk-mxp:
Thanks for adding all the exercises!

Someone to mark the "parallel-letter-frequency" exercise as foregone in `config.json, so we can close this issue?

tomasnorre:
> Every exercise is down now !
> 
> Congratulations to all, specifically to @tomasnorre who did much of it !

It was a team effort. Happy to have contributed.

tomasnorre:
> Thanks for adding all the exercises!
> 
> Someone to mark the "parallel-letter-frequency" exercise as foregone in `config.json`, so we can close this issue?

I don't know how that is done.

homersimpsons:
> Someone to mark the "parallel-letter-frequency" exercise as foregone in config.json, so we can close this issue?

Done in #630

PR #640, opened 2024-02-25
Foregone parallel-letter-frequency exercise

homersimpsons:
PHP is not realy made for multithread. Fibers are new and rarely used.

Completes #630

### DIFF
--- a/config.json
+++ b/config.json
@@ -1236,6 +1236,9 @@
         "prerequisites": [],
         "difficulty": 5
       }
+    ],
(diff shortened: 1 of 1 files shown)


mk-mxp:
Just to be a little pedantic: Fibers are reliably single threaded, too. They are a bit more complicated than Generators, but still not "parallel" in any way or at least "interrupted unexpectedly by some other executed code". I think that *is* a great advantage...

PR #641, opened 2024-02-26
Add ISBN Verifier Exercise

tomasnorre:
I just picked one by "random".


exercises/practice/isbn-verifier/IsbnVerifierTest.php
```diff
@@ -0,0 +1,167 @@
...
+
+    /**
+     * uuid: daad3e58-ce00-4395-8a8e-e3eded1cdc86
+     */
+    public function testCatchInvalidCharactersInOnOtherwiseValidIsbn(): void
+    {
+        $this->assertFalse($this->isbnVerifier->isValid('3598P215088'));
+    }
+
+    public function testInputIsTooLongButContainsAValidIsbn(): void
```
mk-mxp, line 163:
```suggestion
    /**
     * uuid: fb5e48d8-7c03-4bfb-a088-b101df16fdc3
     */
    public function testInputIsTooLongButContainsAValidIsbn(): void
```

mk-mxp (approved):
Just one UUID missing for consistency. Thanks again for the valuable contribution!

exercises/practice/isbn-verifier/IsbnVerifierTest.php
tomasnorre: I must have overlooked that.

mk-mxp:
@ErikSchierboom Do you have an idea, what really went with the CI tests on MacOS? It looks like there is a library missing on the MacOS image. But that was not the case some hours ago. The image tag hasn't changed and the workflow is all the same. Any clues?

ErikSchierboom:
I don't, but what you could do is to pin to a specific version of the Mac runner in .github/workflows/exercise-tests-phpunit-9.yml. It's currently using latest which can cause these sudden breakages

mk-mxp:
Thanks for your patience with CI and me :smile:

tomasnorre:
> Thanks for your patience with CI and me 😄

No worry, we all want to do it as good as possible. Really appreciate your guidance.

PR #646, opened 2024-02-27
CI add back macos test

tomasnorre:
With the failing tests yesterday we were running against `macOS-latest` (12.7.3) but for what ever reason `macOS-13` nor `macOS-14` is tagged as latest, but with the `macOS-13` it works again. 

I'll rebase my open PR when this is merged. 

Technically this shouldn't be needed any more after the `setup-php` as the issue https://github.com/shivammathur/setup-php/issues/823 has been resolved, but I would still opt for updating to `macOS-13` instead of `macOS-12` which `macOS-latest` pointed too. 

If you disagree, this PR can be closed.

### DIFF
--- a/.github/workflows/exercise-tests-phpunit-9.yml
+++ b/.github/workflows/exercise-tests-phpunit-9.yml
@@ -16,7 +16,6 @@ jobs:
       fail-fast: false
       matrix:
         php-version: [8.0, 8.1, 8.2]
-        # os: [ubuntu-22.04, windows-2022, macOS-12]
         os: [ubuntu-22.04, windows-2022, macOS-13]
 
     steps:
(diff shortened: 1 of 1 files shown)


mk-mxp:
If I understand versioning of MacOS correctly, MacOS 12 is an LTS version and thus is prefered (expressed as 'latest') for long term stable CI processes. But I haven't had much time these days to investigate deeper.

Anyways, I would prefer to stop running the tests on MacOS and Windows. Because PHP has become super independent of the underlying OS except for those well documented cases of direct OS proxying. And I believe, but haven't proven yet, that Exercism doesn't and will not teach or use those.

So, thanks for looking into a solution. I'll just leave MacOS deactivated and start a discussion in the forum about that topic.

tomasnorre:
OK. I didn't see that part with LTS I may have overlooked that part.

Issue #629, opened 2024-02-22
Tracking: Missing exercises from problem specifications


mk-mxp:
Directly from `configlet info`.

There are 44 non-deprecated exercises in `exercism/problem-specifications` that are both unimplemented and not in the track config `exercises.foregone` array:

With canonical data:

- [ ] bottle-song
- [ ] circular-buffer
- [ ] complex-numbers
- [ ] custom-set
- [ ] dominoes
- [ ] eliuds-eggs
- [ ] food-chain
- [ ] forth
- [ ] go-counting
- [ ] grep
- [ ] house
- [ ] isbn-verifier
- [ ] killer-sudoku-helper
- [ ] kindergarten-garden
- [ ] knapsack
- [ ] ledger
- [ ] micro-blog
- [ ] ~~parallel-letter-frequency~~ (PHP does not support parallel execution)
- [ ] pov
- [ ] pythagorean-triplet
- [ ] rational-numbers
- [ ] react
- [ ] rectangles
- [ ] rest-api
- [ ] rotational-cipher
- [ ] saddle-points
- [ ] satellite
- [ ] secret-handshake
- [ ] sgf-parsing
- [ ] spiral-matrix
- [ ] square-root
- [ ] state-of-tic-tac-toe
- [ ] strain
- [ ] sublist
- [ ] two-bucket
- [ ] word-search
- [ ] zipper

Without canonical data:

- [ ] dot-dsl
- [ ] error-handling
- [ ] hangman
- [ ] lens-person
- [ ] paasio
- [ ] simple-linked-list
- [ ] tree-building

neenjaw:
I suggest to add these to the forgone exercises:
- the unimplemented sing-song exercises like bottle-song
- math-centric exercises like complex-numbers, rational-numbers, Pythagorean-triplet -- these are niche topics which exclude a lot of users if you aren't already familiar with these math concepts. 
- rest-api -- very old exercise, poorly defined
- zipper -- doesn't make very much sense in a non-functional language.

tomasnorre:
Wauw, I didn't expect it to be this many exercises missing, then there is still work to do :) I Only had the list from #48in24 in mind. 
 
Would it be possible to add a priority to them? High, Medium, Low would be enough I think.

mk-mxp:
> Would it be possible to add a priority to them? High, Medium, Low would be enough I think.

I'm too busy to look at them in more detail at the moment. But that was my plan. And also mark those as _foregone_ that cannot or should not be implemented.

[...]

PR #647, opened 2024-02-27
Add House Exercise

tomasnorre:


exercises/practice/house/.docs/instructions.md
```diff
@@ -0,0 +1,105 @@
+# Instructions
+
+Recite the nursery rhyme 'This is the House that Jack Built'.
+
+> [The] process of placing a phrase of clause within another phrase of clause is called embedding.
```
mk-mxp, line 5:
```suggestion
> The process of placing a phrase of clause within another phrase of clause is called embedding.
```

mk-mxp (changes requested):
Thanks for contribution! Only a small Markdown issue here.

This is one of the ["sing-song" exercises](https://github.com/exercism/php/issues/629#issuecomment-1959560962) @neenjaw suggested to skip. I do not know why, but maybe this should be the last exercise of that kind until we know the reasons. There is enough others to choose.

tomasnorre:
> Thanks for contribution! Only a small Markdown issue here.
> 
> This is one of the ["sing-song" exercises](https://github.com/exercism/php/issues/629#issuecomment-1959560962) @neenjaw suggested to skip. I do not know why, but maybe this should be the last exercise of that kind until we know the reasons. There is enough others to choose.

I didn't think about that before completing it. I'm OK with rejecting it, too.

mk-mxp (approved):
Thanks - one last time for today - for contribution!

PR #649, opened 2024-02-28
Add Micro Blog Exercise

tomasnorre:
Have I misunderstood this exercise? Or is it really this simple? I didn't read any indications that build in functions wasn't allowed, but perhaps it should? Depends on the purpose of the exercise.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1149,6 +1149,14 @@
         "difficulty": 5
       },
       {
+        "slug": "micro-blog",
+        "name": "Micro Blog",
+        "uuid": "203ba680-cdbd-494d-a1b6-af426bc84e59",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1
+      },
+      {
         "slug": "yacht",
         "name": "Yacht",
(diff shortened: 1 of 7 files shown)


mk-mxp:
> Have I misunderstood this exercise? Or is it really this simple?

Yes, it is. Some are really easy and simple, some are really hard.

> I didn't read any indications that build in functions wasn't allowed, but perhaps it should? Depends on the purpose of the exercise.

Instructions say: `The trick to this exercise is to use APIs designed around Unicode characters (codepoints) instead of Unicode codeunits.` And that's `mb_*` instead of `str*` in PHP. If you know them, you're done. If you don't, you learn to know them with this exercise.

mk-mxp (approved):
Nice and straight forward. Did you need ChatGPT for that one, too? :wink: Thanks for contributing!

tomasnorre:
> Nice and straight forward. Did you need ChatGPT for that one, too? 😉 Thanks for contributing!

Haha. No :) Currently I use ChatGPT for generating the tests, and then I try to solve the exercise as TDD. I personally learn more from that. But for this one I managed without any help, and even did the tests without ChatCPT. ;)

PR #651, opened 2024-02-29
Add State of Tic-Tac-Toe Exercise

tomasnorre:


exercises/practice/state-of-tic-tac-toe/StateOfTicTacToe.php
```diff
@@ -0,0 +1,33 @@
...
+ * <link>.
+ *
+ * To disable strict typing, comment out the directive below.
+ */
+
+declare(strict_types=1);
+
+class StateOfTicTacToe
+{
+    public static function gameState(array $board): string
```
mk-mxp, line 29: Please use no `static` methods on the students interface.

mk-mxp (commented):
The current solution using strings as the result is according to specification.

But I think, this is the perfect exercise to use an `enum` as a return type. That would mean, a) changing the implementation and b) pause the merging, until we have PHP8.1 as minimum version.

I would be very happy, if you agree to wait for that. When we merge this now, we cannot change the interface later to use `enum` as that invalidates all submitted solutions.

What do you think?

exercises/practice/state-of-tic-tac-toe/StateOfTicTacToe.php
tomasnorre: My bad, I forgot to clean up. Will remove it tomorrow.

tomasnorre:
> The current solution using strings as the result is according to specification.
> 
> But I think, this is the perfect exercise to use an `enum` as a return type. That would mean, a) changing the implementation and b) pause the merging, until we have PHP8.1 as minimum version.
> 
> I would be very happy, if you agree to wait for that. When we merge this now, we cannot change the interface later to use `enum` as that invalidates all submitted solutions.
> 
> What do you think?

I'm perfectly fine with `enum` and waiting. I think `enum` makes perfectly sense in this case.

tomasnorre:
It's expected to fail in PHP 8.0 as I have already added the `enum` we talked about yesterday. Then we might just need to rebase when we are ready to merge.

mk-mxp:
That's looks great! Looking forward to having a minimum of PHP 8.1...

mk-mxp:
@tomasnorre Today the PHP track got upgraded to the required PHP version. You may rebase this branch and update the PR.

tomasnorre:
I must have done something wrong in the rebasing. As all the changes are included. 

I did

1. Sync my fork
2. git checkout branch
3. git rebase main
4. git push --force

mk-mxp (approved):
Great! Thanks a lot.

PR #654, opened 2024-03-03
Improve contributor experience

### LINKED ISSUE #620: Improve contributor experience
> Can we add any CI job to ensure this formating is kept in next Pull Requests ?

_Originally posted by @homersimpsons in https://github.com/exercism/php/issues/619#issuecomment-1951481621_

- [x] [Ask exercism officials](https://forum.exercism.org/t/add-configlet-fmt-to-ci/9869) to add format checking to configlet workflow using a configuration switch
- [x] Integrate `configlet fmt` into composer scripts (for contributors)
- [x] Add a `composer ci` or such to run all tests with one command
- [x] Add fetching configlet to `composer install` and `composer update` without loading it in CI
- [x] Drop PHP 7.4 from `composer.json`

I think, this would greatly improve the contributor experience.

mk-mxp:
Closes #620 and fixes issues with the GitHub actions workflow for PHPCS linting:

- When a job has a `strategy` key, it must provide a `matrix` in that.
- We don't need the PHPCS PHAR in CI anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,9 @@ composer install # Required dependencies to develop this track
 `bin/configlet` is the official tool for managing Exercism language track repositories.
 See [Building Exercism docs][exercism-configlet].
 
+For convenience, you can use `composer configlet:fmt` to fix formatting issues in the Exercism track files.
+This is included in `composer ci` to run the CI checks locally.
+
 ## Running Unit Test Suite
 
 The tests are run with PHPUnit. A shell loop injecting `exemplar.php` is provided to ease testing.
@@ -35,15 +38,19 @@ Execute the following command to run the tests:
 composer test:run
 ```
 
+This is included in `composer ci` to run the CI checks locally.
+
 ## Running Style Checker
 
-This project use a slightly [modified][local-file-phpcs-config] version of [PSR-12].
+This project uses a slightly [modified][local-file-phpcs-config] version of [PSR-12].
 Use the following commands to apply code style:
 
 ```shell
 composer lint:check # Checks the files against the code style rules
-composer lint:fix # Automatically fix codestyle issues
+composer lint:fix # Automatically fix code style issues
 ```
+
+The `lint:check` is included in `composer ci` to run the CI checks locally.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ The following system dependencies are required:
 
 - `composer`, as recommended in the [PHP track installation docs][exercism-track-installation-composer].
 - [`bash` shell][gnu-bash].
-- PHP V8.2+ CLI
+- PHP V8.2+ CLI.
 - An active Internet connection for installing required tools / composer packages.
 
 Run the following command to get started with this project:
--- a/README.md
+++ b/README.md
@@ -10,19 +10,19 @@ To solve the exercises, head to the [PHP track][exercism-track-home] and check t
 The following system dependencies are required:
 
 - `composer`, as recommended in the [PHP track installation docs][exercism-track-installation-composer].
-- [`bash` shell][gnu-bash]
+- [`bash` shell][gnu-bash].
 - PHP V8.2+ CLI
+- An active Internet connection for installing required tools / composer packages.
 
-Run the following commands to get started with this project:
+Run the following command to get started with this project:
 
 ```shell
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -10,12 +10,37 @@
...
+        "@tests:run"
+    ],
+    "tool:configlet:error": "[ -x bin/configlet ] || ( echo 'Fetch configlet first! Run: \"bin/fetch-configlet\"'; false )",
+    "tool:configlet:fetch": "[ -z \"$CI\" ] || [ -x bin/configlet ] || bin/fetch-configlet",
+    "tool:configlet:update": [
+        "rm bin/configlet || true",
+        "[ -n \"$CI\" ] || bin/fetch-configlet"
+    ],
+    "configlet:fmt": [
+        "@tool:configlet:error",
```
homersimpsons, line 37: Does it make sense to just fetch it here ?

```suggestion
        "@tool:configlet:fetch",
```

composer.json
```diff
@@ -10,12 +10,37 @@
...
+    "post-update-cmd": [
+        "@tool:configlet:update"
+    ],
+    "ci": [
+        "@configlet:fmt",
+        "@lint:check",
+        "@tests:run"
+    ],
+    "tool:configlet:error": "[ -x bin/configlet ] || ( echo 'Fetch configlet first! Run: \"bin/fetch-configlet\"'; false )",
+    "tool:configlet:fetch": "[ -z \"$CI\" ] || [ -x bin/configlet ] || bin/fetch-configlet",
```
homersimpsons, line 31: So we do not want to fetch it in CI. Should we check for the `CI` environment variable or use `--no-scripts` during `composer install` ?

homersimpsons (commented):
I think the change regarding configlet are debatable. But let's do them and see if there is any negative feedback.

README.md
```diff
@@ -10,21 +10,24 @@ To solve the exercises, head to the [PHP track][exercism-track-home] and check t
...
 composer install # Required dependencies to develop this track
 ```
 
 ## Running Exercism resources management
 
-`bin/configlet` is a tool to manage exercism resources in this track.
+`bin/configlet` is the official tool for managing Exercism language track repositories.
 See [Building Exercism docs][exercism-configlet].
 
+For convenience, you can use `composer configlet:fmt` to fix formatting issues in the Exercism track files.
```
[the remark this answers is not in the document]
mk-mxp: Yes, but I wanted to delay that until the test-generator reached MVP status. I will add it now, as you ask for it.

composer.json
```diff
@@ -10,12 +10,37 @@
...
+    "tool:configlet:update": [
+        "rm bin/configlet || true",
+        "[ -n \"$CI\" ] || bin/fetch-configlet"
```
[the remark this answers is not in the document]
mk-mxp: No, the logic of the gate is different. `tool:configlet:fetch` MUST NOT abort with success, when not in CI but install it. But it gives a failure when used in CI. `tool:configlet:update` MUST abort with success when in CI. It took me a while to get that right.

composer.json
mk-mxp: This is a point, where I am not 100% sure. I usually argue, that when you reach this point, you haven't used `composer install` or `... update`, so I see it as an error. Mostly because I cannot rely on the development environment in total. And very often the attempt to fix that causes more problems than it solves.

On the other hand, sometimes this might be enough. But still, when you get there, you did something that was not planned for the automation to do. Must I support this? I don't know...

composer.json
mk-mxp: I usually disable functionality not in total, but for the cases it might fail. Here we don't do anything but fetch configlet, so this might be an option. I'll think about it.

README.md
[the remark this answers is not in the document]
mk-mxp: OK, no, I won't add it now. The problem is handling the `slug` required for exercise generation. That requires some additional scripting, so that only those commands get the slug that need it and the others do not fail.

composer.json
mk-mxp: I think I will cut that down. It looks over-engineered to me now, with all these conditions and checks...

composer.json
mk-mxp: I chose to simply do nothing in case of a missing configlet. The command will just fail for "command not found".

homersimpsons (approved):

PR #655, opened 2024-03-03
CI Add back macOS test

### LINKED ISSUE #645: Re-enable testing on MacOS
Due to an issue out of our control testing on PHP8.1+ with MacOS 12 always fails. Enable it again (I commented it out) when https://github.com/shivammathur/setup-php/issues/823 is fixed.

The issue is fixed, but should we really enable it again? See https://forum.exercism.org/t/php-stop-running-ci-tests-on-macos-and-windows/10025

tomasnorre:
Resolves #645


mk-mxp (approved):
As I could not give you reputation for the test generator, I will give you a lot more for this.

Thanks for your contributions!

Issue #652, opened 2024-03-01
Coordination: Upgrade to PHP 8.1 - 8.3 and PHPUnit 10


mk-mxp:
This is what I think must be done:

- [x] Add PHP 8.3 to the CI of the track and allow it to fail #656 
- [x] Add PHPUnit 10 to the CI of the track and allow it to fail #656 
- [x] Fix the problems that occur with PHP 8.3 and PHPUnit 10
  - [x] Performance degrades when using PHP 8.3 #656 
    - [x] `XDEBUG_MODE=off` solves this issue (XDebug is "more active" in default mode than before)
  - [x] Performance degrades when switching from PHPUnit 9 to 10
    - [x] `XDEBUG_MODE=off` required to get CI passing without timeouts (in #656)
    - [x] Timeouts for "robot-name" exercise during `composer ci` locally. It was not a problem at all with PHPUnit 9.6 (not even a noticable delay). `XDEBUG_MODE=off` doesn't help as much as it did for other exercises. Can be solved by replacing `assertArrayNotHasKey($name, $names, ...` with `assertFalse(isset($names[$name]), ...`. This reduces runtime from minutes to milliseconds. Problem exists with PHPUnit >= 10.4 || < 11 #683 
  - [x] Add a CI test to catch tests that would fail due to the [runtime limit set by Exercism infrastructure](https://exercism.org/docs/building/tooling/test-runners/interface#h-allowed-run-time). Uses `timeout 54s <command>` as a process based limit per exercise (3x 18 seconds, see [forum for calculation details](https://forum.exercism.org/t/test-tracks-for-the-20-seconds-limit-on-test-runners/10536/8)) #683 
- [x] Prepare the [PHP test-runner](https://github.com/exercism/php-test-runner) to upgrade to PHP 8.3 / PHPUnit10. Use `XDEBUG_MODE=off` for performance reasons. https://github.com/exercism/php-test-runner/pull/105
- [x] Prepare the [PHP representer](https://github.com/exercism/php-representer) to upgrade to PHP 8.3. Use `XDEBUG_MODE=off` for performance reasons https://github.com/exercism/php-representer/issues/155
- [x] Do the actual switch to PHP 8.1 - 8.3 and PHPUnit 10 as the production versions #704
  - [x] Merge test runner upgrade PR
  - [x] Update student facing documentation
  - [x] Remove PHP 8.0 / PHPUnit 9 from CI, add PHP8.1/8.2 to PHPUnit10 workflow

[...]

PR #656, opened 2024-03-04
CI Add PHPUnit 10 and PHP 8.3

tomasnorre:
I'm not sure how to add the allow failure on GitHub Actions, with GitLab CI I would know how to do it, but to my knowledge it's not supported in GHA yet. 

This solves, from my understanding, the first 3 bullets in #652 

For some reason, the tests are slower with PHPUnit 10 and PHP 8.3, than PHPUnit 10 and PHP 8.2. I have not dug into that yet, but would be something that would interest me why. 

Todo: 

- [x] Add static DataProviders

### DIFF
--- a/.github/workflows/exercise-tests-phpunit-10.yml
+++ b/.github/workflows/exercise-tests-phpunit-10.yml
@@ -12,11 +12,14 @@ jobs:
   test:
     name: PHP ${{ matrix.php-version }} - ${{ matrix.os }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
+    continue-on-error: true
     strategy:
       fail-fast: false
       matrix:
-        php-version: [8.1, 8.2, 8.3]
-        os: [ubuntu-22.04, windows-2022, macOS-12]
+        #php-version: [8.1, 8.2, 8.3]
+        php-version: [8.3]
+        #os: [ubuntu-22.04, windows-2022, macOS-12]
+        os: [ubuntu-22.04]
 
     steps:
       - name: Set git line endings
--- a/.github/workflows/exercise-tests-phpunit-10.yml
+++ b/.github/workflows/exercise-tests-phpunit-10.yml
@@ -44,4 +44,5 @@ jobs:
         shell: bash
         env:
           PHPUNIT_BIN: 'bin/phpunit-10.phar'
+          XDEBUG_MODE: off
         run: bin/test.sh
--- a/.github/workflows/exercise-tests-phpunit-10.yml
+++ b/.github/workflows/exercise-tests-phpunit-10.yml
@@ -1,0 +1,45 @@
+name: Exercise tests with PHPUnit 10
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+
+jobs:
+  test:
+    name: PHP ${{ matrix.php-version }} - ${{ matrix.os }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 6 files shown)


mk-mxp:
Try: [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error). I also am no GitHub actions expert, but this should do the trick.

Also, please do not run all OSes and PHP versions, yet. `ubuntu-22.04` and PHP 8.3 should be enough. Keep the matrix, maybe comment it out.

And do not update PHP requirement in `composer.json`, that is part of #654 .

.github/workflows/exercise-tests-phpunit-10.yml
```diff
@@ -0,0 +1,48 @@
+name: Exercise tests with PHPUnit 10
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
```
mk-mxp, line 8:
```suggestion
```

We don't have `master` anymore.

.github/workflows/exercise-tests-phpunit-10.yml
```diff
@@ -0,0 +1,48 @@
...
+  pull_request:
+
+jobs:
+  test:
+    name: PHP ${{ matrix.php-version }} - ${{ matrix.os }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
```
mk-mxp, line 15: Isn't `continue-on-error` meant to be in `jobs.test.steps.*`? The one that may fail is the last step.

.github/workflows/exercise-tests-phpunit-10.yml
```diff
@@ -0,0 +1,48 @@
...
+          php-version: ${{ matrix.php-version }}
+          extensions: gmp
+
+      - name: Install dependencies
+        shell: bash
+        run: |
+          curl -Lo ./bin/phpunit-10.phar https://phar.phpunit.de/phpunit-10.phar
+          chmod +x bin/phpunit-10.phar
+
+      - name: Test exercises\
```
mk-mxp, line 44:
```suggestion
      - name: Test exercises
```

mk-mxp (changes requested):

mk-mxp:
About the performance problem: My first guess is XDebug. Maybe `XDEBUG_MODE=off` in the environment of the last job helps?

.github/workflows/exercise-tests-phpunit-10.yml
tomasnorre: As I read the documentation. You can add it like this, to allow all tests to fail. 

But, I see your point, we don't want anything before that to fail. Will adjust.

tomasnorre:
> About the performance problem: My first guess is XDebug. Maybe `XDEBUG_MODE=off` in the environment of the last job helps?

Possible, didn't think about that as I didn't see it activated.

.github/workflows/exercise-tests-phpunit-10.yml
mk-mxp: There are many ways to acheive the goal. I didn't know that using `continue-on-error` is allowed to be used for all jobs. I thought it might be gracefully ignored up there.

I'm fine with any solution doing the trick.

mk-mxp (approved):
OK, performance is now equal to the older versions. Great step forward!

Thanks a lot for the fiddling with the CI.

tomasnorre:
> OK, performance is now equal to the older versions. Great step forward!
> 
> Thanks a lot for the fiddling with the CI.

No Problem. Working as a DevOps, it's not really new to me.

PR #662, opened 2024-03-25
Update introduction.md

niyak93rus:
identity symbol was wrong ('+' instead of '=')

### DIFF
--- a/concepts/arithmetic-operators/introduction.md
+++ b/concepts/arithmetic-operators/introduction.md
@@ -2,7 +2,7 @@
 
 PHP provides a number of operators for performing arithmetic operations. PHP follows the standard mathematical order of operations for its arithmetic operators. The operators that are provided by PHP are:
 
-* Identity (+)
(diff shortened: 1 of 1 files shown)


mk-mxp:
Hi @niyak93rus, welcome to the PHP track!

I understand that you think, the identity operator *must be `=`*. But *Identity* in the arithmetic operator context means a different thing than *Identity* in the comparison operator context. In the context of arithmetic operations *Identity* means "does not change the value". So a literal value `4` *is identical to* `+4`.

Please take a look at the [PHP documentation on arithmetic operators](https://www.php.net/manual/en/language.operators.arithmetic.php) for further details of operator behaviour.

mk-mxp:
Closed, because there is no failure in the document.

niyak93rus:
Oh, man, I feel a bit stupid now. Actually, not a bit lol. Thanks for clarifying, @mk-mxp

PR #663, opened 2024-03-25
Re-write test generator with templates and unit tests

mk-mxp:
I decided to re-write the test generator based on template files instead of `nikic/php-parser`. It had too many limitations. And I didn't want to dig into another huge PHP generation library. Producing the code is not the main problem here.

While learning about the data structure behind the canonical data, I understood that the basic problem is not writing the PHP code but interpreting the data. Based on that idea, showing everything unknown to the user of the generator is key to informed usage.

So I developed a tree of data structures for the known data using TDD. Rendering these structures using templates simplified the process of dumping the code to the file. I documented all that in `contribution/documentation/generator`.

It now works for all remaining practice exercises, but does not produce the students file, yet.

What do you think about this approach, @homersimpsons @tomasnorre ?


homersimpsons:
I won't get time to review this before this weekend (31st March).

I know nothing about plantuml files though.

I really think it would be more test-able using twig templating and flysystem memory storage for tests.

mk-mxp:
@homersimpsons Don't worry about PlantUML, it is only some Code-To-Diagram application I know. The SVGs in the documentation are made from those files, there is nothing more to it.

I chose plain text files for templates and PHP `sprintf()` to fill in stuff. It's very basic.

homersimpsons:
> @homersimpsons Don't worry about PlantUML, it is only some Code-To-Diagram application I know. The SVGs in the documentation are made from those files, there is nothing more to it.


Wouldn't mermaid diagram works?

> I chose plain text files for templates and PHP `sprintf()` to fill in stuff. It's very basic.

Okay, but we may be limited to use some loops or conditions.

tomasnorre:
I'll not find time to read this before after Easter.

IsaacG:
+cc @exercism/php A generator update would be swell!

mk-mxp:
@IsaacG You are welcome to lean in. I have at least 3 attempts from the last years lying around here but none is useful for more than a handful of exercises. I'm happy with using any other tracks generator for this track, too, as long as we don't have to maintain huge tooling.

IsaacG:
@mk-mxp What's worked well for a number of other tracks is a per-exercise template file and a generator.
I've got a Python/JinJa2 generator I used largely unchanged across a handful of tracks that might be worth pulling in here. 

Given PHP, having a PHP based generator would probably be nicer, with a per-exercise PHP file ... but if you want something pre-built that already works, the Python one could get the job done. Thoughts on using Python + JinJa to generate the tests here?

homersimpsons:
> @mk-mxp What's worked well for a number of other tracks is a per-exercise template file and a generator.

This is what I have in #674, using twig.

IsaacG:
> This is what I have in #674, using twig.

TBH this looks fairly promising to me and in line with what many other tracks have.

mk-mxp:
@homersimpsons @IsaacG Please feel free to take this forward. I would prefer using plain PHP templates (as PHP itself is a template language anyways), but I don't see this as a must have.

I do not have time or energy to move this forward. Whoever comes up with a test generator usable for reproducing the existing exercises from canonical data that also is integrated into docs, workflows and maybe PHP beginner friendly - you will have done the track a service.

IsaacG:
@homersimpsons Do you want to revisit #674 and get that launched? Using PHP directly sounds nice ... but if Twig is easier and gets the job done, then that's a win. A working generator is much better than no generator. If you need help writing generators for additional exercises after the first one or two are done, I might be able to help out.

PR #592, opened 2023-12-26
Update MeetupTest.php

beqqrry-aws:
The PR to add strict types to all PHP files (https://github.com/exercism/php/pull/383) broke the tests for this assignment because the months are all sent as strings. This PR fixes the tests by changing the months to ints.

PR #665, opened 2024-03-26
Fix Meetup exercise tests

mk-mxp:
- Test with `int` as specified in students file stub.
- This requires adjusting the example solution to show it can still be done.
- Also remove superficial strict types file headers.

Follow-up of #592 (closed by creator)

### DIFF
--- a/exercises/practice/meetup/.meta/example.php
+++ b/exercises/practice/meetup/.meta/example.php
@@ -1,32 +1,10 @@
 <?php
 
-/*
- * By adding type hints and enabling strict type checking, code can become
- * easier to read, self-documenting and reduce the number of potential bugs.
- * By default, type declarations are non-strict, which means they will attempt
- * to change the original type to match the type specified by the
- * type-declaration.
- *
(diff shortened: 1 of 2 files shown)


mk-mxp:
This change would invalidate a bunch of solutions, how can we prevent re-running tests for too old solutions? Or shouldn't we?

homersimpsons:
> This change would invalidate a bunch of solutions, how can we prevent re-running tests for too old solutions? Or shouldn't we?

I'm not sure how we should proceed about this, @ErikSchierboom do you have any suggestion?

As this aligns the type to the student file stub, to me we should proceed with this change. It's been 3y this stub is like this, so I wouldn't expect this to break too much student code.

This also aligns with the canonical-data.json from problem specifications.

homersimpsons (approved):

mk-mxp:
I just realized that this is the featured exercise for `#48in24` week 32 and PHP will be one of the 3 languages of that week. So I decided to re-run all tests on community solutions to get as much "working code" there as possible.

PR #683, opened 2024-04-14
Test timeout issues

mk-mxp:
## PHPUnit 10.5 `assertArrayNotHasKey()`

PHPUnit `^10.3` has an issue with `assertArrayNotHasKey()`. Only this (and the opposite `assertArrayHasKey()`) are affected. The issue is not listed in PHPUnit's issue list / mentioned in Pull Requests, but the issue is gone in `11.0.0`.

This contains a workaround for this and enables upgrading to PHPUnit 10.5. I replaced that assertion with a functionally equivalent one, but that prints some misleading output in addition to the helpful message. So it should be reverted when upgrading to PHPUnit 11.

## CI limit for test runtime

I also added a time limit of 18 (lowest value possible on my machine with equal performance to Online Test Runner) x 3 (to compensate lower CI performance) = 54 seconds for each exercise test class using `timeout` command. This should prevent running into performance degradation issues in the future.

As [discussed in the forum](https://forum.exercism.org/t/test-tracks-for-the-20-seconds-limit-on-test-runners/10536) there is no recommendation or guiding data for such a limit. Only guesses about "test runner is expected to be slower than GitHub CI". But the production limit of 20 seconds exists and is enforced. So I took a guess [based on my measurements](https://forum.exercism.org/t/test-tracks-for-the-20-seconds-limit-on-test-runners/10536/8).

Part of #652

### DIFF
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -58,7 +58,7 @@ function test {
     cp "${exercise_dir}/.meta/${example_file}" "${outdir}/${exercise_file}.${file_ext}"
   fi
 
-  eval "${PHPUNIT_BIN}" --default-time-limit 7 --enforce-time-limit --fail-on-risky --no-configuration "${outdir}/${test_file}"
+  eval "${PHPUNIT_BIN}" --default-time-limit 20 --enforce-time-limit --fail-on-risky --no-configuration "${outdir}/${test_file}"
 }
 
 function installed {
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -58,7 +58,7 @@ function test {
     cp "${exercise_dir}/.meta/${example_file}" "${outdir}/${exercise_file}.${file_ext}"
   fi
 
-  eval "${PHPUNIT_BIN}" --no-configuration "${outdir}/${test_file}"
+  eval "${PHPUNIT_BIN}" --default-time-limit 7 --enforce-time-limit --fail-on-risky --no-configuration "${outdir}/${test_file}"
 }
 
 function installed {
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -58,7 +58,23 @@ function test {
     cp "${exercise_dir}/.meta/${example_file}" "${outdir}/${exercise_file}.${file_ext}"
   fi
 
-  eval "${PHPUNIT_BIN}" --default-time-limit 20 --enforce-time-limit --fail-on-risky --no-configuration "${outdir}/${test_file}"
+  # The time-limit `18` is based on an approximation of the performance.
+  # Online Test Runner is appr. 3x faster than GitHub CI on Ubuntu.
+  #
+  # The total runtime limit of the Online Test Runner is 20 seconds including
+  # Docker container launch and processing the results. That leaves appr. 18
+  # seconds for all tests of an exercise. But that's not the `18` here.
+  #
+  # The longest running exercise tests in CI are in `alphametics`. It has 3 slow
+  # test cases (9 letters and 10 letters) that require appr. 15 secs each in CI
+  # and appr. 5 seconds each in the Online Test Runner.
+  #
+  # Putting that together gives a max. of 6 seconds for each of the slow test
+  # cases in the Online Test Runner before hitting the timeout in production
+  # -> use 3x 6secs = 18 seconds as a CI limit to ensure tests can actually
+  # be run in production.
+  # See: https://forum.exercism.org/t/test-tracks-for-the-20-seconds-limit-on-test-runners/10536/8
+  eval "${PHPUNIT_BIN}" --default-time-limit 18 --enforce-time-limit --fail-on-risky --no-configuration "${outdir}/${test_file}"
 }
 
 function installed {
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -58,16 +58,17 @@ function test {
     cp "${exercise_dir}/.meta/${example_file}" "${outdir}/${exercise_file}.${file_ext}"
   fi
 
-  # The time-limit `54` is based on an approximation of the performance.
-  # Online Test Runner is appr. 3x faster than GitHub CI on Ubuntu.
+  # `54s` timeout is an approximation to ensure the tests will not timeont in Exercism Test Runner.
   #
-  # The total runtime limit of the Online Test Runner is 20 seconds including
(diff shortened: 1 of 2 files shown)


mk-mxp:
@homersimpsons Sorry, I thought I had figured that out right. But I somehow haven't remembered things correctly...

mk-mxp:
@homersimpsons Another attempt using measured data now. I put the data into the forum, and a longish comment into the test script.

exercises/practice/robot-name/RobotNameTest.php
```diff
@@ -73,26 +73,42 @@ public function testResetName(): void
         $this->assertMatchesRegularExpression('/\w{2}\d{3}/', $name2);
     }
 
+    /**
+     * PHPUnit ^10.5 has an issue with
```
homersimpsons, line 77: Do we have any link to backup this? I could not find such issue.

How hard is it to switch to PHPUnit ^11?

bin/test.sh
```diff
@@ -58,7 +58,23 @@ function test {
...
+  #
+  # The longest running exercise tests in CI are in `alphametics`. It has 3 slow
+  # test cases (9 letters and 10 letters) that require appr. 15 secs each in CI
+  # and appr. 5 seconds each in the Online Test Runner.
+  #
+  # Putting that together gives a max. of 6 seconds for each of the slow test
+  # cases in the Online Test Runner before hitting the timeout in production
+  # -> use 3x 6secs = 18 seconds as a CI limit to ensure tests can actually
+  # be run in production.
+  # See: https://forum.exercism.org/t/test-tracks-for-the-20-seconds-limit-on-test-runners/10536/8
```
homersimpsons, line 76: Thank you for all this investigation!

I'm wondering if we need that much details here. Maybe it is sufficient to say something along:
"Exercism gives 20s for the test runner to complete. It is better to report a timeout ourselves so the student can see which tests are slow. We limit it to 18s to account for setup and report."

Also it may be possible to play with the `@small` and the like.

exercises/practice/robot-name/RobotNameTest.php
mk-mxp: I could not find any issue or merge request in PHPUnit repos, too. I have isolated the problem to experiment with versions. Then I found out, that it was introduced in 10.3.0 together with a massive change in internal event propagation. A lot of that was changed for V11, but 10.3 developed forward to 10.5 and still has this issue. So no, there is no link.

PHPUnit 11 does not support PHP 8.1. So I think it is not an option now. Exercism supports 3 minor versions and that should stay. We will, I hope, upgrade in early 2025.

bin/test.sh
mk-mxp: This is no student facing limit. It is a limit to warn us, when updates / upgrades of PHP or PHPUnit degrade performance the same as we experienced now. Without the change in `robot-name`, CI would have passed all tests but students would never ever have a chnace to pass the tests in the online test runner. It also helps detect, when new exercises contain stuff, PHP cannot calculate fast enough to pass the tests (`alphametics` is already at the limit).

I wanted to have this level of detail there to ensure, every contributor can understand that removing / raising this limit is no solution to make CI pass. And when CI speed changes (GitHub will upgrade CI runners one day), contributors must know how to adjust the limit.

I was thinking about `@large` for the slow tests. But I didn't want to have a limit only for things we already know are slow. This limitation must work for all tests, no matter how they perform now. And for all future tests, too.

exercises/practice/robot-name/RobotNameTest.php
homersimpsons: >  I have isolated the problem to experiment with versions. Then I found out, that it was introduced in 10.3.0 together with a massive change in internal event propagation.

Okay, thank you for the investigation still...

bin/test.sh
homersimpsons: Oh sorry, I do not know why but I thought we were in the test runner.

Okay, it makes sense to have those details here then. But i'm not sure that this is exactly what we want. In fact, the `default-time-limit` put a constraint on a single test (https://github.com/sebastianbergmann/phpunit/blob/877f44b74d2563366f76e044d3fd710ebb94aeca/src/Framework/TestRunner.php#L435) so if we add a new test to `alphametics` in the near future we may still encounter timeout issues.

I cannot seem to find a way to add a timeout to a group of test or to a testsuite.

Maybe we should time the whole `eval` instead to ensure it is less than 18s.

mk-mxp: I know that PHPUnit command line limits are per test method, but had no other solution at hand. `timeout` is a better way to go, didn't know of that command.

I will make that work in Alpine...Thanks!

bin/test.sh
```diff
@@ -58,7 +58,17 @@ function test {
...
+  # The time-limit `54` is based on an approximation of the performance.
+  # Online Test Runner is appr. 3x faster than GitHub CI on Ubuntu.
+  #
+  # The total runtime limit of the Online Test Runner is 20 seconds including
+  # Docker container launch and processing the results. That leaves appr. 18
+  # seconds for all tests of an exercise.
+  #
+  # Putting that together gives a max. of 18 seconds x3 = 54 seconds for any
+  # test class in CI.
+  # See: https://forum.exercism.org/t/test-tracks-for-the-20-seconds-limit-on-test-runners/10536/8
```
homersimpsons, line 70: I think "Exercism Test Runner" would be more precise than "Online Test Runner", I also suggest the following comment, I think it is clearer:

```suggestion
  # `54s` timeout is an approximation to ensure the tests will not timeont in Exercism Test Runner.
  #
  # 1. Exercism Test Runner is around 3 times faster than GitHub CI on Ubuntu.
  #    See: https://forum.exercism.org/t/test-tracks-for-the-20-seconds-limit-on-test-runners/10536/8
  # 2. Exercism Test Runner should complete in 20s and involves:
  #    - Starting Docker container (~1s)
  #    - Running the test suite
  #    - Processing the results (~1s)
  #
  # This gives 18s maximum for the test suite to run in the Exercism Test Runner.
  # Hence the test suite should complete in less than 18s x 3 = 54s in GitHub CI on Ubuntu.
```

The current comment still is okay, so feel free to update it or no (or even suggest another version).

homersimpsons (approved):
Thank you for your investigation. I'm proposing a new wording for the comment about the timeout. But I still approve the pull request as-is.

PR #694, opened 2024-04-20
Sync All your base

### LINKED ISSUE #678: #48in24: all-your-base 2024-04-30
- [x] Run `bin/configlet sync -u -e all-your-base` (updates the Markdown files and maybe `tests.toml`)
  - docs: instructions unsynced: all-your-base
  - docs: introduction missing: all-your-base

Replaces #688

mk-mxp:
This will be the `#48in24` exercise on 2024-04-30. It is marked as ready to sync in the [forum](http://forum.exercism.org/t/48in24-exercise-04-30-all-your-base/10613). Important files changed, re-running tests on submitted student code is required.

- Some test inputs differed from the problem spec, but did not change expected behaviour.
- Students interface design expected `null` as return value for errors. This is changed to exceptions to align with modern PHP track interface design.
- Also some test (zeros and leading zeros) behaved different from problem spec expectations. As these values are not mentioned in instructions and so have no defined behaviour, we chose to conform to problem specs (return `[0]`).

Closes #678

### DIFF
--- a/exercises/practice/all-your-base/AllYourBaseTest.php
+++ b/exercises/practice/all-your-base/AllYourBaseTest.php
@@ -214,8 +214,6 @@ class AllYourBaseTest extends PHPUnit\Framework\TestCase
     }
 
     /**
-     * We cannot know which condition comes first. So no message is enforced.
-     *
      * uuid 0e6c895d-8a5d-4868-a345-309d094cfe8d
      * @testdox Both bases are negative
      */
--- a/exercises/practice/all-your-base/AllYourBaseTest.php
+++ b/exercises/practice/all-your-base/AllYourBaseTest.php
@@ -3,8 +3,12 @@
 declare(strict_types=1);
 
 /**
- * These tests are adapted from the canonical data in the
- * `problem-specifications` repository.
+ * Design choices (standing against problem specs):
+ * - Return `null` for errors
+ * - Zero and leading zeros are errors. Problem spec says, these become `[0]`
+ *
+ * Because of these choices, some inputs for error cases differ from problem
+ * specs, but target the same conditions as problem specs.
  */
 class AllYourBaseTest extends PHPUnit\Framework\TestCase
 {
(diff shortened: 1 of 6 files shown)


homersimpsons:
> @homersimpsons What do you think? Should I invalidate existing solutions to align with problem spec? This would open the way for Exceptions (InvalidArgumentException), too.

I would vote in favour of invalidating existing solutions to align with problem specifications.

Advantages (of aligning):
- 48in24 will certainly bring some new solutions that would help to fill community solutions with correct ones
- Updating in the future may be worse
- It will certainly be easier to sync later

Inconveniences (of aligning):
- students that want to check existing community solutions may not find what they were looking for
- some students will see their solutions broken

But I would prefer another point of view from @ErikSchierboom, what do you think about aligning the tests with the spec and invalidating existing solutions ? Maybe there is a correct way to approach this like a message or something to nudge the students to update their solutions and point them in the correct direction (I know they will at least see the diff though).

ErikSchierboom (approved):

homersimpsons:
@ErikSchierboom looks like there was a misunderstanding. The current changes are not aligning the test behaviour with canonical data, it is not throwing exception and won't invalidate existing solutions.

Do you approve the changes as they are now and reject the alignment of the tests ?

ErikSchierboom:
Ah, I misunderstood. I'll have to take another look Tuesday then

mk-mxp:
To put in my opinion: I'm very much for updating and modernizing this exercise (and all others), sync-ing to problem specs as much as possible and throwing exceptions as is done in modern PHP everywhere. I read a lot about not invalidating existing solutions, as they are the knowledge base to lookup help / improvements. And yet, how helpful are solutions that follow outdated practices? So to me, updating is the way to go. But I am quite as unsure as @homersimpsons if that's the Exercism way to handle that.

exercises/practice/all-your-base/AllYourBaseTest.php
```diff
@@ -1,32 +1,14 @@
...
 
 /**
- * These tests are adapted from the canonical data in the
- * `problem-specifications` repository.
+ * Design choices (standing against problem specs):
+ * - Return `null` for errors
+ * - Zero and leading zeros are errors. Problem spec says, these become `[0]`
+ *
+ * Because of these choices, some inputs for error cases differ from problem
+ * specs, but target the same conditions as problem specs.
```
ErikSchierboom, line 11: I don't think this is information that should be in the test files, as students won't know anything about canonical data (as a matter of fact, I think the existing comment is also unhelpful).

If you'd like to document design choices, you can add them to a `.meta/design.md` file.

exercises/practice/all-your-base/AllYourBaseTest.php
```diff
@@ -35,108 +17,196 @@ public static function setUpBeforeClass(): void
         require_once 'AllYourBase.php';
     }
 
+    /**
+     * uuid 5ce422f9-7a4b-4f44-ad29-49c67cb32d2c
```
ErikSchierboom, line 21: What is the goal of adding the uuid?

ErikSchierboom (commented):
If I'm seeing this correctly, this PR doesn't break any existing solutions, right?

exercises/practice/all-your-base/AllYourBaseTest.php
mk-mxp: There is more than that use case. It is the first and only file occational contributors look into when "fixing" tests. It may be their first hint that tests are not unique to this place. And it is the place where experienced PHP programmers expect such information about test design.

Additionally, we have little `design.md` files that are up-to-date or even adapted to PHP completely. Which means, even regular contributors don't care for other files than the tests. So these comments are, in my eyes, the best way to add relevant decisions to the track.

Anyways, I hope we don't need this comment, because we agree to align with problem spec completely.

exercises/practice/all-your-base/AllYourBaseTest.php
mk-mxp: We currently have no good way to document the relation between test case and problem specification. The test names are no reliable source of truth, the test cases sometimes differ in details. So this is a) machine readable and b) a clear sign: this is in sync with problem specs.

mk-mxp:
> If I'm seeing this correctly, this PR doesn't break any existing solutions, right?

@ErikSchierboom Yes, I didn't break anything, yet. But I would like to do so. Changing behaviour for zero and trailing zeros would invalidate all solutions, as we treat that as error now (while problem spec wants `[0]`). Using exceptions instead of `null` would invalidate all solutions, too. Shall I break nothing or everything?

exercises/practice/all-your-base/AllYourBaseTest.php
ErikSchierboom: > It is the first and only file occational contributors look into when "fixing" tests.

I understand, but we should be optimizing for students, not contributors and as a student, I would find that a _very_ confusing comment.

exercises/practice/all-your-base/AllYourBaseTest.php
ErikSchierboom: > The test names are no reliable source of truth

Why are they not a reliable source of truth? Test names hardly ever change.

ErikSchierboom:
I've checked how many solutions there are for this exercise, and it is not a ton so I'm fine with changing the exercise to be more idiomatic. I'm not a fan of the maintainer/contributor-specific changes this PR introduces which might confuse the student.

homersimpsons (approved):
Looks better like this, thank you!

PR #701, opened 2024-05-03
Sync minesweeper

### LINKED ISSUE #692: #48in24: minesweeper 2024-05-14
- [x] Run `bin/configlet sync -u -e minesweeper` (updates the Markdown files and maybe `tests.toml`)
  - docs: instructions unsynced: minesweeper
  - docs: introduction missing: minesweeper
  - metadata: unsynced: minesweeper

Replaces #686

mk-mxp:
As discussed in #692 this is a complete makeover. Re-running tests for all solutions required.

- The students interface is now a class with a method, expecting the input to the constructor. This is based on the thinking, that the minesweeper board is a more long living entity that might have more behaviour than only knowing the annotated version. E.g. it may as well be used to verify moves (flagging mines) against the board.
- Tests are ordered as in problem spec. This is not the best order for TDD, though.

Closes #692


homersimpsons:
@mk-mxp Thanks for this contribution.

> The students interface is now a class with a method, expecting the input to the constructor. This is based on the thinking, that the minesweeper board is a more long living entity that might have more behaviour than only knowing the annotated version. E.g. it may as well be used to verify moves (flagging mines) against the board.

This approach kind of makes sense. It looks like [java has this approach](https://github.com/exercism/java/blob/ce492756d4bc0d00410b4df00e19882b5634aeb2/exercises/practice/minesweeper/src/main/java/MinesweeperBoard.java), [cpp is just exposing a "static" method in the class](https://github.com/exercism/cpp/blob/main/exercises/practice/minesweeper/.meta/example.cpp) and other languages ([javascript](https://github.com/exercism/javascript/blob/main/exercises/practice/minesweeper/minesweeper.js), [python](https://github.com/exercism/python/blob/main/exercises/practice/minesweeper/minesweeper.py) and [rust](https://github.com/exercism/rust/blob/main/exercises/practice/minesweeper/src/lib.rs)) are just exposing a function.

The tests looks okay.

Just a feedback about the `example.php`, I know this is just a proof and not an exemplar solution, but I feel like it could be simplified in case any student read it by removing the "caching" of the annotated version.

Also maybe we could remove some indirections to get something easier to follow:
```php
private function countMinesAround(int $row, int $col): int
{
    $mineCount = 0;
    $mineCount += $this->isMine($row - 1, $col - 1) ? 1 : 0;
    $mineCount += $this->isMine($row - 1, $col    ) ? 1 : 0;
    $mineCount += $this->isMine($row - 1, $col + 1) ? 1 : 0;
    $mineCount += $this->isMine($row    , $col - 1) ? 1 : 0;
    $mineCount += $this->isMine($row    , $col + 1) ? 1 : 0;
    $mineCount += $this->isMine($row + 1, $col - 1) ? 1 : 0;
    $mineCount += $this->isMine($row + 1, $col    ) ? 1 : 0;
    $mineCount += $this->isMine($row + 1, $col + 1) ? 1 : 0;

    return $mineCount;
}

private function isMine(int $row, int $col): bool
{
    return isset($this->minefield[$row][$col]) && $this->minefield[$row][$col] === '*';
}
```

This may be a bit less performant but more readable.

I am still approving the changes as-is though.

homersimpsons (approved):

mk-mxp:
I'd really love to have a better means of discussing code than this asynchronous PR thing...

**TL;DR**: There is more than one perspective on "easier to follow" or "readable".

PS: I skipped much code below to keep the text shorter. I hope it still is understandable.

> it could be simplified [...] by removing the "caching" of the annotated version.

One of the things I really got used to is [Memoization](https://en.wikipedia.org/wiki/Memoization). I just have to advertise it whenever it makes sense. Like here.

> maybe we could remove some indirections to get something easier to follow:
> 
> ```
> private function countMinesAround(int $row, int $col): int
> [...]
> ```

The first thing I do not think is "more readable" is `$this->isMine() ? 1 : 0`. This is repeated many times and without identifying the spreadout parts (line beginning and end with varying things inbetween) one doesn't see this as a meaningful concept. As it is repeated anyways, it should be named `mineScore($row, $col)` (or such) naming that concept.

```php
    $mineCount += $this->mineScore($row - 1, $col - 1);
    $mineCount += $this->mineScore($row - 1, $col    );
[...]
    $mineCount += $this->mineScore($row + 1, $col + 1);
```

The next thing is the 2 written out loops with that pattern breaking missing center. One thing that comes into my mind immediately is: That must fail at the borders. Why is there no exception for the edges? The answer is hidden in `isMine()`, which takes into account that there are out-of-bounds situations. So I need to dig deep before understanding this.

You made `isMine()` to be something very different. It is now `isOnBoardAndIsMine()`, which might be renamed to `countsAsMine()` or such. But this then would not name the rules for being on the board or those for being a mine. So it would take 3 methods: `isOnBoard()`, `isMine()` and `countsAsMine()` to make this clear without in depth thinking about why a square might not exist.

So, maybe it is more understandable to exclude the borders in the first place. Meaning, when the square is not on the board, do not try to score it as a mine. That makes it an `if ($row > 0)` around the first 3 occurences of `$mineCount += $this->mineScore($row - 1, ...);` and `if ($row < \count($this->minefield))` for the last 3 `$mineCount += $this->mineScore($row + 1, ...);`. The columns then also need such an `if()`.

I think you can see, that this leads to exactly my solution...

Is your solution or my solution "more readable to students"? I think, it depends on the learning type of the student. Global learners need to see the whole context at once and then follow connections from concept to concept (like your solution shows them). Sequential learners need to see one concept only, then the next concept, until finally the whole context is explored and concepts are connected (like my solution shows them). So, it depends.

homersimpsons:
> I'd really love to have a better means of discussing code than this asynchronous PR thing...

asynchronous discussion is fine for me, but we can discuss this elsewhere.

> One of the things I really got used to is [Memoization](https://en.wikipedia.org/wiki/Memoization). I just have to advertise it whenever it makes sense. Like here

Yes but I do not think this is the point for this exercise

> The first thing I do not think is "more readable" is $this->isMine() ? 1 : 0. This is repeated many times and without identifying the spreadout parts (line beginning and end with varying things inbetween) one doesn't see this as a meaningful concept. As it is repeated anyways, it should be named mineScore($row, $col) (or such) naming that concept.

This could be an improvement too in fact, I didn't want to go too far from your current solution

> One thing that comes into my mind immediately is: That must fail at the borders. Why is there no exception for the edges? The answer is hidden in `isMine()`, which takes into account that there are out-of-bounds situations. So I need to dig deep before understanding this.

As you said it is the first thing you think about and reading the next function gives you the answer. While there are a lot of lines to read from `countMinesAround` and `countMinesOfRowAroundCol` and both depend on the fact that a bound check was done before calling the function. A user will certainly skip reading with the bounds check complexity.

> You made `isMine()` to be something very different. It is now `isOnBoardAndIsMine()`, which might be renamed to `countsAsMine()` or such. But this then would not name the rules for being on the board or those for being a mine. So it would take 3 methods: `isOnBoard()`, `isMine()` and `countsAsMine()` to make this clear without in depth thinking about why a square might not exist.

`isOnBoardAndIsMine()` might be sufficient, that is just naming

> Is your solution or my solution "more readable to students"? I think, it depends on the learning type of the student. Global learners need to see the whole context at once and then follow connections from concept to concept (like your solution shows them). Sequential learners need to see one concept only, then the next concept, until finally the whole context is explored and concepts are connected (like my solution shows them). So, it depends.

I kind of agree with that. From my point of view in one case the user will have to follow:
- We count the mines around
- We count the mines of a the row above (unless it is out of board)
  - To count the mines of a row we count the mine on the left, middle and right column (checking for out of bound columns and assuming row is not out of bound). We will even count the center that was obviously not a mine from the step 0 (for loops)
- We count the mines of the middle row
- We count the mines of the bottow row (with a bound check)

While on the other hand the user will certainly read as:
- "Oh there is lot of calls to the same function, I can certainly assume we are checking every cell around"
- "But what about borders?"
- "Okay bound check is done as part of the `isMine()` function"

That said I do not have any material to rely on to assert that one method is easier to read / understand than the other.
As this code is not really for students I did not push this forward more.

PR #724, opened 2024-05-27
Sync space age

### LINKED ISSUE #713: #48in24: space-age 2024-06-04
- [x] Run `bin/configlet sync -u -e space-age` (updates the Markdown files and maybe `tests.toml`)
  - docs: instructions unsynced: space-age
  - space-age: missing 1 test case 

mk-mxp:
This will be the `#48in24` exercise on 2024-06-04. Important files changed, but re-running tests on submitted student code not required.

- Documentation sync'ed and checked for mismatches to track tests
- Tests sync'ed with the problem specs
- I did not add the test for a wrong planet, as that makes all existing solutions invalid
- Removed the test for internal storage of `seconds`, as that should not be tested for

Closes #713

### DIFF
--- a/exercises/practice/space-age/.docs/instructions.md
+++ b/exercises/practice/space-age/.docs/instructions.md
@@ -1,18 +1,28 @@
 # Instructions
 
-Given an age in seconds, calculate how old someone would be on:
+Given an age in seconds, calculate how old someone would be on a planet in our Solar System.
 
-   - Mercury: orbital period 0.2408467 Earth years
-   - Venus: orbital period 0.61519726 Earth years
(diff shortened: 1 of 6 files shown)


homersimpsons (approved):
Everything looks good! Thank you!

mk-mxp:
> About the "wrong planet test"

The whole exercise is centered around floating point numbers and very easy. Having exceptions seems odd to me, as we specify all planets as methods explicitly. There are more advanced exercises, where adding such a test makes more sense.

I think that test is for approaches using "planet name to data structure / method / function mapping in student land", where one should actually have such a test. With the complete interface and all methods pre-declared - why should anyone try to use `__call()` to implement everything dynamically? That would be the only way I see to make use of an exception for a wrong planet.

PR #736, opened 2024-06-08
Sync matching-brackets

### LINKED ISSUE #714: #48in24: matching-brackets 2024-06-18
- [x] Run `bin/configlet sync -u -e matching-brackets --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
  - docs: instructions unsynced: matching-brackets
  - matching-brackets: missing 3 test cases

mk-mxp:
This will be the `#48in24` exercise on 2024-06-18. Important files changed, re-running tests on submitted student code required.

- 7 tests from problem spec were added. This may cause students solutions to fail, as cases were not covered before. I found [1 community solution](https://exercism.org/tracks/php/exercises/matching-brackets/solutions/caseysj) failed (I tested about 20 of the 253 finished solutions according to track status).
- The function name of the students interface is not changed. Problem spec suggests `isPaired`, we use `bracket_match`. I think that's OK.

Closes #714

### DIFF
--- a/exercises/practice/matching-brackets/.docs/instructions.md
+++ b/exercises/practice/matching-brackets/.docs/instructions.md
@@ -1,5 +1,5 @@
 # Instructions
 
-Given a string containing brackets `[]`, braces `{}`, parentheses `()`,
-or any combination thereof, verify that any and all pairs are matched
-and nested correctly.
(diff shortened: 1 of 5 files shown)


homersimpsons (approved):
About the function name let's keep it like this to not break existing solutions. I think it is okay to break some solutions (looks like there are not a lot of broken solution) by adding those test.

Thank you!

Issue #684, opened 2024-04-15
Tracking: Syncing exercises for #48in24


mk-mxp:
For `#48in24` exercises should be synced a few days before they are featured. Exercism core team polishes these exercises some time in advance (see [`#48in24` forum threads](https://forum.exercism.org/c/exercism/48in24/364)), but there is no announcement of being ready, yet. So I think, we should **not** work through this list as fast as possible, only sync them a maximum of 2 weeks ahead of the dates.

Most of these exercises also appear in #628.

To sync an exercise with problem specifications, take these steps:

- [x] Run `bin/configlet sync -u -e <slug> --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
- [x] Drop strict types comments from test file and example code (these are useless)
- [x] Add and sync test meta data to tests (`uuid` / `@testdox` in DocBlocks)
- [x] Decide on adding / adjusting / ordering test cases to match current problem specs

Do not redesign the student's interface or add test cases that would invalidate existing community solutions. These are extra tasks, which should be discussed in advance.

`Week | Featured at | exercise slug`

- [x] W43 | 2024-11-05 | eliuds-eggs #822
- [x] W44 | 2024-11-12 | collatz-conjecture #823
- [x] W45 | 2024-11-19 | run-length-encoding #824
- [x] W46 | 2024-11-26 | armstrong-numbers #825

- [x] W47 | 2024-12-03 | diamond #826
- [x] W48 | 2024-12-10 | largest-series-product #827

Exercises that are synced can be removed from these lists at will.

mk-mxp:
When #48in24 started, the Exercism staff worked on exercise instructions to make them "more inclusive and engaging". That was the reason to **not** work on as many exercises as possible, but wait until about 14 days before the featured date.

Now Jeremy and Erik (the former staff) are no longer working full time for Exercism. And they worked through all of #48in24 to have that done before "leaving" (they are still active as volunteers).

So I think we can now update all the remaining #48in24 exercises without having outdated instructions / extra work for synchronising again. Still we should synchronize them before their feature date.

FYI @sarad1p1ty @fejan-malek @tomasnorre @homersimpsons

mk-mxp:
@sarad1p1ty @fejan-malek @tomasnorre @homersimpsons

Looks like we are all done now. Thanks to all of you for contributing! You helped improving the 48 exercises featured in #48in24, and raised the quality experienced by thousands of students every year.

[...]

PR #738, opened 2024-06-11
change: sync binary search

fejan92:
Completed all the todos which is mentioned here.

https://github.com/exercism/php/issues/731

Also added 2 new missing test cases from problem_specification.

### DIFF
--- a/exercises/practice/binary-search/.docs/instructions.md
+++ b/exercises/practice/binary-search/.docs/instructions.md
@@ -11,7 +11,7 @@ Binary search only works when a list has been sorted.
 
 The algorithm looks like this:
 
-- Find the middle element of a *sorted* list and compare it with the item we're looking for.
+- Find the middle element of a _sorted_ list and compare it with the item we're looking for.
 - If the middle element is our item, then we're done!
 - If the middle element is greater than our item, we can eliminate that element and all the elements **after** it.
 - If the middle element is less than our item, we can eliminate that element and all the elements **before** it.
--- a/exercises/practice/binary-search/.meta/example.php
+++ b/exercises/practice/binary-search/.meta/example.php
@@ -1,27 +1,5 @@
 <?php
 
-/*
- * By adding type hints and enabling strict type checking, code can become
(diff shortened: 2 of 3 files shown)


mk-mxp (approved):
Thank you very much, great work again!

Please wait before doing the next sync task. About 2 weeks before the date in the issue heading is the right time to sync the exercise. Exercism staff updates problem specifications also. So too fast means syncing again.

Happy to see a PR for #732 "spiral-matrix" after 2024-06-24!

fejan92:
Yes sure I will take  "spiral-matrix" as a priority after 2024-06-24!

Till then any issue in priority?

mk-mxp:
Priority? No, nothing with priority.

We have a list of exercises in issue #684 (at the end, starting with `W1 | 2024-01-16 | leap`) where syncing is required, but they already were part of the `#48in24` challenge. These can be done at any time. I can create issues for each exercise, if you want to do these.

Do you have advanced knowledge of PHP and PHPUnit? There are still some practice exercises missing (see #629). This also requires more knowledge of `configlet` and the inner workings of Exercism (https://exercism.org/docs/building/tracks/practice-exercises). Before adding one, open a forum discussion in Programming/PHP to discuss which one and how the students interface should look like.

fejan92:
> Please wait before doing the next sync task. About 2 weeks before the date in the issue heading is the right time to sync the exercise. Exercism staff updates problem specifications also. So too fast means syncing again.

You told me to wait till 2 weeks correct so I can't contribute to any sync task till that time.

> Do you have advanced knowledge of PHP and PHPUnit? There are still some practice exercises missing (see https://github.com/exercism/php/issues/629). This also requires more knowledge of configlet and the inner workings of Exercism

Yes, I am aware of advanced of PHP and even work on frameworks of PHP. And about the PHP unit, I am still getting my hands on it.

So as you suggested me issue is that we have to create a missing exercise is that correct?

fejan92:
If you allow me I can work on re-sync of exercise from W1.

Issue #739, opened 2024-06-12
#48in24: Sync leap


mk-mxp:
- [x] Run `bin/configlet sync -u -e leap --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
- [x] Drop strict types comments from test file and example code (these are useless)
- [x] Add and sync test meta data to tests (`uuid` / `@testdox` in DocBlocks)
- [x] Decide on adding / adjusting / ordering test cases to match current problem specs

Do not redesign the student's interface or add test cases that would invalidate existing community solutions. These are extra tasks, which should be discussed in advance.

mk-mxp:
@fejan92 You may start here... :smiley: Happy Coding :computer:

fejan92:
Sure soon I will work on this.
Thanks!

PR #748, opened 2024-06-13
Sync leap

fejan92:
https://github.com/exercism/php/issues/739

There is no change after running the command `bin/configlet sync -u -e leap --yes --docs --filepaths --metadata --tests include`

Test cases function name is updated with canonical JSON file and also added missing test cases.


mk-mxp (approved):
Nice work! Thanks for contributing.

PR #754, opened 2024-06-15
extra minor patch

Eual11:
minor changes for sake of mathematical correctness xD


mk-mxp:
@Eual11 Thanks for wanting to contribute. We intentionally **don't use** the circles area here. It is not logical, to cover a whole pizza with a gap-less monolithic layer of cheese. That's why the instructions say: "to use the following formula" and not the area of the pizza.

PR #756, opened 2024-06-18
Sync roman-numerals

fejan-malek:
https://github.com/exercism/php/issues/742

@mk-mxp 

I have removed strict type from the example and test cases file.

I also added `testdox` and `UUID` please take note that no new test cases are added and I removed 2 test cases that are not present in the `test.toml` file.

This below 2 functions are removed.
`test1998 , test2999`

### DIFF
--- a/exercises/practice/roman-numerals/.meta/example.php
+++ b/exercises/practice/roman-numerals/.meta/example.php
@@ -1,27 +1,5 @@
 <?php
 
-/*
- * By adding type hints and enabling strict type checking, code can become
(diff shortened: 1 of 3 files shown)


mk-mxp (changes requested):
Please add the new tests from problem specifications. They do not add new behaviour (unicode in `reverse-string` was new behaviour). So we should have them.

If you choose not to add them, mark them with `include=false` in `tests.toml`.

mk-mxp (approved):
Great work! Thanks a lot for improving this.

PR #757, opened 2024-06-22
Sync list-ops

fejan-malek:
https://github.com/exercism/php/issues/744


mk-mxp (approved):
One of the very new exercises. Thanks for doing the boring work, too!

Issue #743, opened 2024-06-12
#48in24: Sync protein-translation


mk-mxp:
- [x] Run `bin/configlet sync -u -e protein-translation --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
  - docs: instructions unsynced: protein-translation
  - metadata: unsynced: protein-translation
- [x] Drop strict types comments from test file and example code (these are useless)
- [x] Add and sync test meta data to tests (`uuid` / `@testdox` in DocBlocks)
- [x] Decide on adding / adjusting / ordering test cases to match current problem specs
  - protein-translation: missing 30 test cases
  - Missing `tests.toml` file!

Do not redesign the student's interface or add test cases that would invalidate existing community solutions. These are extra tasks, which should be discussed in advance.

mk-mxp:
No new tests, but data provider spread out into single test methods.

Issue #755, opened 2024-06-17
Discussion: Use of data providers in test classes


mk-mxp:
Using data providers is a convenient way to reduce code and increase coherence inside the test classes. Having the test case data separated from the test code itself follows best practices and is encouraged in professional use.

But in Exercism it raises a bunch of problems:

- In the test runner, data provider based tests must be handled separately
  - Nested test suites issued for data provider tests only
  - Code reflection based line number calculation fails with input variables being on separate lines
  - Test values are only included in students test feedback when tests are designed accordingly
  - See https://github.com/exercism/php-test-runner/issues/116 for some issues
- PHPUnit converts values to a different format (not like PHP code) to emit them in test names
- For beginners, even when working with the CLI, the test data is not intuitively connected to the test code
- It is much more inconvenient to connect UUIDs of practice exercises to test values in a data provider
- In the test generator, data providers might add more complexity and so may not be used

So I would prefer to convert existing data provider based tests to flat test methods and not add new exercises with data providers.

Any opinions about that?

tomasnorre:
As there are limited tests, even thought they are more or less always looking the same. I'm OK with not using dataproviders, if that improves user experience in Online Editor. 

I would always reach for data providers personally in a private project, with descriptive keys.

But I think user experience is key here. We will not need to tweak tests that often.

mk-mxp:
Still using DataProvider:

- [ ] ~~binary~~ (this is deprecated)
- [x] #1019
- [x] #973

PR #759, opened 2024-06-22
Sync protein-translation

fejan-malek:
https://github.com/exercism/php/issues/743

### DIFF
--- a/exercises/practice/protein-translation/ProteinTranslationTest.php
+++ b/exercises/practice/protein-translation/ProteinTranslationTest.php
@@ -1,27 +1,5 @@
 <?php
 
-/*
- * By adding type hints and enabling strict type checking, code can become
- * easier to read, self-documenting and reduce the number of potential bugs.
- * By default, type declarations are non-strict, which means they will attempt
- * to change the original type to match the type specified by the
- * type-declaration.
- *
- * In other words, if you pass a string to a function requiring a float,
- * it will attempt to convert the string value to a float.
- *
- * To enable strict mode, a single declare directive must be placed at the top
- * of the file.
- * This means that the strictness of typing is configured on a per-file basis.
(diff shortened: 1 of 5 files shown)


exercises/practice/protein-translation/ProteinTranslationTest.php
```diff
@@ -174,6 +256,18 @@ public function testTranslationStopsIfStopCodonInMiddleOfSixCodonSequence(): voi
         );
     }
 
+    /**
+     * @testdox Invalid codon throws exception
+     * uuid 6b3d2d8a-6f6e-4f6e-8a0b-0d7d8a0b0d8a
+     * @dataProvider invalidCodonDataProvider
```
mk-mxp, line 262: Please make these test cases separate tests. See this issue for reasons: https://github.com/exercism/php/issues/755

Do not implement `uuid 1e75ea2a-f907-4994-ae5c-118632a1cb0f`, because it is replaced by `uuid 9eac93f3-627a-4c90-8653-6d0a0595bc6f` (see `tests.toml` and `"reimplements"` key in `canonical-data.json`.

If you want to read more about how this is connected, see https://github.com/exercism/problem-specifications/#changing-tests

mk-mxp (changes requested):
That's a nice piece of work. Thank you!

exercises/practice/protein-translation/ProteinTranslationTest.php
fejan-malek: Yes, I noticed the `reimplements` in the `test.toml` file also in another exercise. So if there is a `reimplements` keyword then we don't have to implement such test cases is that correct?

mk-mxp: That's correct. Thank you!

mk-mxp (approved):

PR #770, opened 2024-07-06
Sync circular buffer

fejan-malek:
https://github.com/exercism/php/issues/746


mk-mxp (approved):
Thanks for helping with this!

Issue #783, opened 2024-07-27
Re-design `robot-simulator`


mk-mxp:
The current exercise design has many problems:

- Students must implement constants for directions with no TDD style guidance - since PHP8 this could be an enum
- The test accesses `position` and `direction` properties directly, but `protected` fields are provided in the stub
- The class is not named like the file - `Robot` vs. `RobotSimulator.php`
- ~~The tests expect methods to move the robot, while problem spec only has `instructions`~~ (will be solved in #784) 
- ~~The tests are clumped together with no clear separation between test cases~~

Problem specifications do not ask for a certain implementation. They define the problem statement as open as possible. Our design also should allow as many solutions as possible.

Current stats:

```text
Robot Simulator

Started 672
Attempts 4,498 (avg. 6.7)
Completions 446 (66.4%)
Mentoring requests 44 (6.5%)
```

## No discussion about these changes

- When we break things, rename the class to match the file name.
- [x] Test no extra methods to move the robot - `instructions()` only or fluent interface only. (see #784)
- [x] Structure tests according to problem spec - Nested test cases share the test name prefix of the group. Each test case is a test method.

## Directions

Suggestion 1: Strings only

This leaves all possibilities open. Mentors can suggest `enum` or such to students. But `enum` is optimal for interfaces, internal `enum` usage just complicates things.

Suggestion 2: Provide `enum` stub

In addition to the tests from problem specifications we test the `enum` implementation first. When that is implemented, the actual problem is tested for. In these tests we must use the `enum` type, too (like the constants now).

This forces students to use `enum`. It raises the bar for students not familiar with that. But `enum` would be a good choice in a real life scenario.

Suggestion 3: Provide ready-made `enum`

As before, this forces students to use an `enum` type. But the level of required knowledge is lower (only usage is required). But still, it makes no sense to work with other types internally then.

Which way to go? Other suggestions?

## Accessing state

I think testing the state by accessing fields was done to force students to implement a state encapsulating solution with `__get()` or such. The exercise is marked as `topic: ['oop']` in `config.json`, which hints on why it may have been designed like this.

To stay as close to the problem spec as possible, the OOP-isms are seemingly superfluous. Accessing protected properties via `__get()` could be avoided OOP-style also.

Suggestion 1: `public` properties

This is closest to the problem specification and the easiest to implement. But it violates fundamental rules of OOP - hide the details behind methods.

Suggestion 2: `get` methods

[...]

Issue #781, opened 2024-07-27
#48in24: robot-simulator 2024-08-06


mk-mxp:
- [x] Run `bin/configlet sync -u -e robot-simulator --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
  - docs: instructions unsynced: robot-simulator
- [x] Drop strict types comments from test file and example code (these are useless)
- [x] Add and sync test meta data to tests (`uuid` / `@testdox` in DocBlocks)
- [x] Decide on adding / adjusting / ordering test cases to match current problem specs

Do not redesign the student's interface or add test cases that would invalidate existing community solutions. These are extra tasks, which should be discussed in advance.

mk-mxp:
This is designed in a way we cannot actually annotate it with the problem spec UUIDs. So I do not sync and annotate the tests, but request re-designing it in #783.

PR #784, opened 2024-07-27
Sync robot-simulator

mk-mxp:
Sync'ed according to #781. Changes tests heavily in structure, ~~but not~~ **and** in content. So ~~no~~ **re-testing** against existing solutions required.

- Spread out test cases to individual test methods
- Test only from the interface specified in problem specifications
- ~~Still~~ Re-design as discussed in #783 

I did find a way to do a re-structuring without altering the interface relied on by existing solutions. There are now no extra methods used, that I think were older than the current `instructions` property in problem spec.

The exercise is marked as `topic: ['oop']` in `config.json`, which hints on why there were specialised methods to change the internal state. And why our `example.php` (and an additional test) does method chaining (fluent interface design).

To stay as close to the problem spec as possible, the OOP-isms are seemingly superfluous. A fluent interface design deviates totally from the string style `instructions`. On the other hand one could transform the string input into the fluent interface while writing the tests, and also allow immutable instances as a more modern approach that way. But still, I think such OOP-isms should only be in additional PHP track exercises.

### DIFF
--- a/exercises/practice/robot-simulator/.meta/example.php
+++ b/exercises/practice/robot-simulator/.meta/example.php
@@ -100,6 +100,16 @@ class RobotSimulator
         return $this;
     }
 
+    public function getPosition(): array
+    {
+        return $this->position;
+    }
+
+    public function getDirection(): string
+    {
+        return $this->direction;
+    }
+
     /**
      * List all possible clockwise turn combinations
      * @return array
--- a/exercises/practice/robot-simulator/.meta/example.php
+++ b/exercises/practice/robot-simulator/.meta/example.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-class Robot
+class RobotSimulator
 {
     public const DIRECTION_NORTH = 'north';
     public const DIRECTION_EAST = 'east';
@@ -41,9 +41,9 @@ class Robot
 
     /**
      * Turn the Robot clockwise
-     * @return Robot
+     * @return RobotSimulator
      */
-    public function turnRight(): \Robot
+    public function turnRight(): \RobotSimulator
     {
         $this->direction = self::listDirectionsClockwize()[$this->direction];
         return $this;
@@ -51,9 +51,9 @@ class Robot
 
     /**
      * Turn the Robot counterclockwise
-     * @return Robot
+     * @return RobotSimulator
      */
-    public function turnLeft(): \Robot
+    public function turnLeft(): \RobotSimulator
     {
         $this->direction = self::listDirectionsCounterClockwize()[$this->direction];
         return $this;
@@ -61,9 +61,9 @@ class Robot
 
     /**
      * Advance the Robot one step forward
-     * @return Robot
+     * @return RobotSimulator
      */
-    public function advance(): \Robot
+    public function advance(): \RobotSimulator
     {
         switch ($this->direction) {
             case self::DIRECTION_NORTH:
--- a/exercises/practice/robot-simulator/.meta/example.php
+++ b/exercises/practice/robot-simulator/.meta/example.php
@@ -16,17 +16,18 @@ class RobotSimulator
     ) {
     }
 
-    /**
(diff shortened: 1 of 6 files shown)


mk-mxp:
@fejan-malek I cannot add you as a reviewer, because you are no maintainer yet. Anyways, what do you think about a fluent interface here?

fejan-malek:
@mk-mxp I agree with using a fluent interface because it helps write clean code, better code readability, and reduces code complexity.

What do you think about the existing solution already submitted in this exercise? Is it gonna affect these changes?

mk-mxp:
> What do you think about the existing solution already submitted in this exercise? Is it gonna affect these changes?

@fejan-malek Not using an implemented method does not invalidate any existing solution.

mk-mxp:
I've been trying out several variants for a fluent interface. The problem is, that all the benefits of the fluent interface are hidden in the tests. The student experiences nothing but "additional work to return `$this`".

The exercise becomes rather "boring", with only conditionally adjusting values. "Boring implementation" can be a quality of good design, but is discouraging for students at Exercism. There would be only one challenge left in the exercise. The string splitting and looping disappears.

These two points made me decide: The exercise shall stay as close as possible to problem specifications. Fluent interfaces are a concept to introduce in an extra exercise, most likely a PHP track only exercise.

homersimpsons:
Note that I didn't took time to take a look yet, I will probably look into this tomorrow.

About fluent interfaces I'm not in favour of having those: https://ocramius.github.io/blog/fluent-interfaces-are-evil/

That said it could be great to introduce them in an exercise just so the students know they can write code like this and understand how to read code like this.

exercises/practice/robot-simulator/RobotSimulator.php
```diff
@@ -27,34 +27,19 @@
...
+    protected array $position;
 
-    /**
-     *
-     * @var string
-     */
-    protected $direction;
+    protected string $direction;
 
     public function __construct(array $position, string $direction)
```
homersimpsons, line 36: Should we add `@param` annotation for the `$position` parameter?

exercises/practice/robot-simulator/RobotSimulator.php
```diff
@@ -27,34 +27,19 @@
...
-        throw new \BadMethodCallException("Implement the turnRight method");
-    }
-
-    public function turnLeft(): self
-    {
-        throw new \BadMethodCallException("Implement the turnLeft method");
-    }
-
-    public function advance(): self
+    public function instructions(): void
```
homersimpsons, line 41: Should we provide the parameter? or let the student figure it out?

homersimpsons (commented):
Thanks for the changes. I will take a look into #783, I think it would make sense to think about the re-design first so newly published solutions will remain correct in the short future.

exercises/practice/robot-simulator/.meta/example.php
```diff
@@ -2,68 +2,68 @@
...
+        }
 
-    /**
-     *
-     * @var string
-     */
-    protected $direction;
+        foreach ($this->mapInstructionsToActions($instructions) as $action) {
+            $this->$action();
+        }
```
homersimpsons, line 30: (non-blocking) This keeps the idea of the previous example. But I feel like having dynamic function call is more complex than needed.

I suggest to inline the mapping with the function call:

```suggestion
        foreach (str_split($instructions) as $instruction) {
            match ($instruction) {
              'A' => $this->advance(),
              'L' => $this->turnLeft(),
              'R' => $this->turnRight(),
            }
        }
```

homersimpsons (approved):
Just a small comment, but I'm okay with the current changes. Thank you for the changes !

exercises/practice/robot-simulator/.meta/example.php
mk-mxp: Thanks for requesting this. Also used `match()` in `advance()`, as it is much more concise and understandable.

PR #798, opened 2024-09-06
Update instructions.append.md

abdelrahman-gado:
Add semicolon for anonymous function in the example


mk-mxp (approved):
Thanks for reporting and fixing this!

PR #803, opened 2024-09-10
sync phone-number

tomasnorre:
Solves issue #800

### DIFF
--- a/exercises/practice/phone-number/.docs/instructions.md
+++ b/exercises/practice/phone-number/.docs/instructions.md
@@ -2,21 +2,26 @@
 
 Clean up user-entered phone numbers so that they can be sent SMS messages.
 
-The **North American Numbering Plan (NANP)** is a telephone numbering system used by many countries in North America like the United States, Canada or Bermuda. All NANP-countries share the same international country code: `1`.
+The **North American Numbering Plan (NANP)** is a telephone numbering system used by many countries in North America like the United States, Canada or Bermuda.
+All NANP-countries share the same international country code: `1`.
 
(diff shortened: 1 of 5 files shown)


tomasnorre:
Long time no see. I have started this PR. But would like some feedback if I have understood the tasks correctly. 

I have done following: 
Ran `bin/configlet sync -u -e phone-number --yes --docs --filepaths --metadata --tests include`
Added uuid to test cases
Remove strict types comment in example code and tests file. 

Testcases with uuid: 2de74156-f646-42b5-8638-0ef1d8b58bc2 and 4a1509b7-8953-4eec-981b-c483358ff531 are not implemented, as they would break current community solutions. So how to deal with those? 

I don't see which 4 tests cases that aren't implemented according to https://github.com/exercism/problem-specifications/blob/main/exercises/phone-number/canonical-data.json . 
A pointer would be nice.

mk-mxp:
Thank you for adjusting the DocBlocks. Don't forget to un-draft the PR, so it can be merged.

For compatibility remove testing for the exception message or use `expectExceptionMessageMatches()` with a RegExp that can handle both texts. I'd prefer not testing for the message.

tomasnorre:
> For compatibility remove testing for the exception message or use `expectExceptionMessageMatches()` with a RegExp that can handle both texts. I'd prefer not testing for the message.

Will do one of the next days. Don't have time right now. But will take care asap.

mk-mxp (approved):
Thanks for your work! I think `expectExceptionMessageMatches()` needs the RegExp to be surrounded by limiters (e.g. `/.* than 11 digits/`). But I always have to try that out myself :smiley:

PR #808, opened 2024-09-24
Sync: Two Bucket

tomasnorre:
Resolves: #807 

I have added the "setup" to each function, but there are still some diffs, to the `arrange, act, assert`. 

Which one do you prefer? or do we go for a 3rd option?

One: 
```php
public function testMeasureUsingBucketOneOfSize3AndBucketTwoOfSize5StartWithBucketOne(): void
{
    $buckOne = 3;
    $buckTwo = 5;
    $goal = 1;
    $starterBuck = 'one';
 
    $subject = new TwoBucket();
    $solution = $subject->solve($buckOne, $buckTwo, $goal, $starterBuck);
 
    $this->assertEquals(4, $solution->numberOfActions);
    $this->assertEquals('one', $solution->nameOfBucketWithDesiredLiters);
    $this->assertEquals(5, $solution->litersLeftInOtherBucket);
}
```
Two: 
```php
public function testMeasureOneStepUsingBucketOneOfSize1AndBucketTwoOfSize3StartWithBucketTwo(): void
{
    $subject = new TwoBucket();
    $solution = $subject->solve(1, 3, 3, 'two');

    $this->assertEquals(1, $solution->numberOfActions);
    $this->assertEquals('two', $solution->nameOfBucketWithDesiredLiters);
    $this->assertEquals(0, $solution->litersLeftInOtherBucket);
}
```

As there are 3 assertions, I don't know if you prefer having 

```php
$expectedNumberOfActions = 1;
$expectedBucketName = 'two';
$expectedLitersInOtherBucket = 0;
```

Or keep the assertions as is.

### DIFF
--- a/exercises/practice/two-bucket/TwoBucketTest.php
+++ b/exercises/practice/two-bucket/TwoBucketTest.php
@@ -17,13 +17,8 @@ class TwoBucketTest extends PHPUnit\Framework\TestCase
      */
     public function testMeasureUsingBucketOneOfSize3AndBucketTwoOfSize5StartWithBucketOne(): void
     {
-        $buckOne = 3;
-        $buckTwo = 5;
-        $goal = 1;
-        $starterBuck = 'one';
-
         $subject = new TwoBucket();
-        $solution = $subject->solve($buckOne, $buckTwo, $goal, $starterBuck);
+        $solution = $subject->solve(3, 5, 1, 'one');
 
         $this->assertEquals(4, $solution->numberOfActions);
         $this->assertEquals('one', $solution->nameOfBucketWithDesiredLiters);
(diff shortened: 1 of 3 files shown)


exercises/practice/two-bucket/TwoBucketTest.php
```diff
@@ -106,38 +114,26 @@ public function testMeasureUsingBucketOneOfSize2AndBucketTwoOfSize3StartWithBuck
...
-        $this->twoBucket->solve($buckOne, $buckTwo, 5, 'one');
+        $subject = new TwoBucket();
+        $subject->solve($buckOne, $buckTwo, 5, 'one');
     }
 
     /**
      * uuid: aac38b7a-77f4-4d62-9b91-8846d533b054
+     * @testdox With the same buckets but a different goal, then it is possible
      */
     public function testReachabilityNotPossibleToReachGoalStartWithBucketOneAndEndWithBucketTwo(): void
```
mk-mxp, line 133:
```suggestion
    public function testWithSameBucketsButDifferentGoalItIsPossible(): void
```

mk-mxp (changes requested):
Wow, thank you for your quick action!

I'd go with option 2, without extra `expected variables`. We currently have a mix of both input options, and all input parameters are short enough to easily fit into one line. Same for the output, using variables or an array doesn't make it better.

mk-mxp (approved):
Thanks a lot for your contribution!

PR #815, opened 2024-10-07
sync bank-account

### LINKED ISSUE #812: #48in24: bank-account 2024-10-22
- [x] Run `bin/configlet sync -u -e bank-account --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
  - docs: instructions unsynced: bank-account
  - metadata: unsynced: bank-account
  - bank-account: missing 17 test cases
  - Most likely: the `tests.toml` file is missing, but the tests are there

tomasnorre:
Resolves #812 

The last test case is not implemented as it's requires concurrent calls. 
https://github.com/exercism/problem-specifications/blob/main/exercises/bank-account/canonical-data.json#L345

### DIFF
--- a/exercises/practice/bank-account/.meta/tests.toml
+++ b/exercises/practice/bank-account/.meta/tests.toml
@@ -1,0 +1,61 @@
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
+
+[983a1528-4ceb-45e5-8257-8ce01aceb5ed]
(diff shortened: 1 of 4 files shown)


exercises/practice/bank-account/.meta/tests.toml
```diff
@@ -0,0 +1,61 @@
...
+description = "Cannot withdraw more than deposited"
+
+[4f381ef8-10ef-4507-8e1d-0631ecc8ee72]
+description = "Cannot withdraw negative"
+
+[d45df9ea-1db0-47f3-b18c-d365db49d938]
+description = "Cannot deposit negative"
+
+[ba0c1e0b-0f00-416f-8097-a7dfc97871ff]
+description = "Can handle concurrent transactions"
```
mk-mxp, line 61:
```suggestion
description = "Can handle concurrent transactions"
include = false
comment = PHP does not support concurrency (on a level we expect students to work in)
```

mk-mxp (changes requested):
Thanks for your work! I'll take a closer look at the changes these days, time runs out now...

mk-mxp (approved):
Very nice! Thanks a lot for contributing.

PR #832, opened 2024-10-30
sync diamond

### LINKED ISSUE #826: #48in24: diamond 2024-12-03
- [x] Run `bin/configlet sync -u -e diamond --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
  - docs: instructions unsynced: diamond
  - metadata: unsynced: diamond
- [ ] Decide on adding / adjusting / ordering test cases to match current problem specs

tomasnorre:
Resolves #826

### DIFF
--- a/exercises/practice/diamond/.docs/instructions.md
+++ b/exercises/practice/diamond/.docs/instructions.md
@@ -1,22 +1,21 @@
 # Instructions
 
-The diamond kata takes as its input a letter, and outputs it in a diamond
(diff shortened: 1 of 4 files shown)


tomasnorre:
This is not completely done, as I wasn't aware how to deal with the change of function `diamond()` to `rows()` according to the https://github.com/exercism/problem-specifications/blob/main/exercises/diamond/canonical-data.json

PR #834, opened 2024-10-31
Deprecate deprecated exercises from problem specifications

### LINKED ISSUE #813: Ensure deprecated exercises are deprecated
This [forum post](http://forum.exercism.org/t/binary-expected-output-is-not-decimal-representation-of-a-number/13281/2) says that Binary is deprecated for very long, but PHP still has it active.

As deprecating exercises does not do any harm to existing solutions and people currently attempting it, we should ensure, that all deprecated exercises from problem-specifications are marked as deprecated and so not shown to new users.

I'm sure, `configlet` has a command to list those exercises and they can easily be marked deprecated in the central `config.json`.

tomasnorre:
Resolves #813 

I know the `checkDeprecatedExercises` isn't super nice, but it was the fastest way I knew to get inform about which exercises was deprecated. 

I can look into the `configlet` repository, to see if I can incorporate the functionally there. 

Output from script: 

```php
shell> php checkDeprecatedExercises.php
array(5) {
  [0]=>
  string(10) "accumulate"
  [1]=>
  string(9) "beer-song"
  [2]=>
  string(6) "binary"
  [3]=>
  string(15) "scale-generator"
  [4]=>
  string(7) "trinary"
}
```

### DIFF
--- a/bin/checkDeprecatedExercises.php
+++ b/bin/checkDeprecatedExercises.php
@@ -1,0 +1,162 @@
+<?php
+
+/*
+ * By adding type hints and enabling strict type checking, code can become
+ * easier to read, self-documenting and reduce the number of potential bugs.
+ * By default, type declarations are non-strict, which means they will attempt
+ * to change the original type to match the type specified by the
+ * type-declaration.
+ *
+ * In other words, if you pass a string to a function requiring a float,
+ * it will attempt to convert the string value to a float.
+ *
+ * To enable strict mode, a single declare directive must be placed at the top
+ * of the file.
+ * This means that the strictness of typing is configured on a per-file basis.
+ * This directive not only affects the type declarations of parameters, but also
+ * a function's return type.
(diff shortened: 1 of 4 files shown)


mk-mxp:
There's nothing wrong with the checking file, I think. I only think about the location. Maybe put it into `contribution/`? Or actually make it executable (adding a `#/usr/bin/env php` line to the front, setting `+x` file mode)?

Adding it to the [PHP CodeSniffer exclusion list here](https://github.com/exercism/php/blob/8257e2bcd90b0ccb9682b6b8a69fd897d1928053/phpcs.xml#L50) should remove the strict types header requirement.

mk-mxp (approved):
Looks good to me. Thanks for solving this problem!

tomasnorre:
> Looks good to me. Thanks for solving this problem!

You're welcome. Only happy to help.

PR #835, opened 2024-10-31
Fix: Various spellchecks

astyanax:
I used various automated tools to determine mostly orthographic errors, but I edited whatever errors I could find.
I manually fixed everything (no automated edits).
I'm thinking that the "translater" change was an overkill, but I wanted to be pedantic.
edit: I added the "[no important files changed]" bit in the commit message

### DIFF
--- a/exercises/practice/protein-translation/ProteinTranslationTest.php
+++ b/exercises/practice/protein-translation/ProteinTranslationTest.php
@@ -200,7 +200,7 @@ class ProteinTranslationTest extends PHPUnit\Framework\TestCase
      * uuid d0f295df-fb70-425c-946c-ec2ec185388e
      * @testdox Translate RNA strand into correct protein list
      */
-    public function testTranslatoRnaStrandToCorrectProteinList(): void
+    public function testTranslateRnaStrandIntoCorrectProteinList(): void
     {
         $this->assertEquals(
             ['Methionine', 'Phenylalanine', 'Tryptophan'],
--- a/exercises/practice/protein-translation/ProteinTranslationTest.php
+++ b/exercises/practice/protein-translation/ProteinTranslationTest.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 class ProteinTranslationTest extends PHPUnit\Framework\TestCase
 {
-    private ProteinTranslation $translater;
+    private ProteinTranslation $translator;
 
     public static function setUpBeforeClass(): void
     {
@@ -13,7 +13,7 @@ class ProteinTranslationTest extends PHPUnit\Framework\TestCase
 
     public function setUp(): void
     {
-        $this->translater = new ProteinTranslation();
+        $this->translator = new ProteinTranslation();
     }
 
     /**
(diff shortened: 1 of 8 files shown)


tomasnorre:
Thanks for you PR. As a lot of the spelling that you have correct come from https://github.com/exercism/problem-specifications , it would be great if you could create a PR against, that repository too. 

This will make sure that all tracks on exercism.org will profit from this change next time their exercises gets synced.

homersimpsons:
> This PR changes the API of some exercises, which will result in already completed exercises will fail. Therefor this can in my opinion not be merged, sorry. 
> 
> @mk-mxp, @neenjaw or @homersimpsons please way in here. 

Looking at the changes it seems that only the `ProteinTranslationTest.php` could impact the results (others are markdown files). But reading the change it is just a renaming of a class variable used within the test. This should have no effect on test outcome.

(I'm on mobile so maybe I didn't see another change)

About the suggestion to update problem-specifications repository I agree with that if those typos are there too.

tomasnorre:
> Looking at the changes it seems that only the `ProteinTranslationTest.php` could impact the results (others are markdown files). But reading the change it is just a renaming of a class variable used within the test. This should have no effect on test outcome.

I think you're right, I overlooked that the only changes done was in PHP was the test file. Shouldn't break anything.

exercises/practice/protein-translation/ProteinTranslationTest.php
```diff
@@ -193,18 +193,18 @@ public function testToCodonsTranslateToProteins(): void
...
-        $this->assertEquals(['Leucine', 'Leucine'], $this->translater->getProteins('UUAUUG'));
+        $this->assertEquals(['Leucine', 'Leucine'], $this->translator->getProteins('UUAUUG'));
     }
 
     /**
      * uuid d0f295df-fb70-425c-946c-ec2ec185388e
      * @testdox Translate RNA strand into correct protein list
      */
-    public function testTranslateRnaStrandToCorrectProteinList(): void
+    public function testTranslatoRnaStrandToCorrectProteinList(): void
```
mk-mxp, line 203:
```suggestion
    public function testTranslateRnaStrandIntoCorrectProteinList(): void
```

mk-mxp (changes requested):
Thanks a lot for these corrections! Just one small change required, where Search&Replace seems to have gone too far.

exercises/practice/protein-translation/ProteinTranslationTest.php
tasxatzial, line 203: This should be reverted, as the new name has some issues.

tasxatzial: Dismiss my suggestion, as it's overridden by mk-mxp's. I hadn't seen the other reviews before posting my comment.

astyanax:
Thank you all for your comments and careful consideration, and thank you Tomas for the patch!

p.s. It'll take some time to grok `problem-specifications` and how `configlet` interacts with it, [#track-data-tooling](https://github.com/exercism/problem-specifications?tab=readme-ov-file#track-data-tooling) implies that corrections would need to be explicitly applied/overridden.

mk-mxp (approved):
Thanks everyone for contributing!

PR #849, opened 2024-11-08
ci: Allow for single exercise test exection

tomasnorre:
I often want to execute only one of the tests in the exercises when adjusting tests, to speed up the process. 

This PR will allow me to e.g. run only the tests for exercise `book-store` by adding an additional parameter.

It can be done in more way. 

```sh
composer test:run -- book-store
```

```sh
PHPUNIT_BIN='phpunit' bin/test.sh book-store 
```

I hope this will get accepted as it gives me a better workflow when updating the PHP track.

### DIFF
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -13,18 +13,12 @@ file_ext="php"
 function main {
   has_failures=0
 
-  if [ $# -ge 1 ] && [ -n "$1" ]; then
-    specific_exercise=$(find ./exercises/{practice,concept} -name $1 -type d | sort)
-    for exercise_dir in $specific_exercise; do
-        test "${exercise_dir}" "example"
-        if [[ $? -ne 0 ]]; then
-          has_failures=1
-        fi
-      done
-      return $has_failures
+  name_filter=()
+  if [ -n "$1" ]; then
+    name_filter=("-name" "$1")
   fi
 
-  all_practice_exercise_dirs=$(find ./exercises/practice -maxdepth 1 -mindepth 1 -type d | sort)
+  all_practice_exercise_dirs=$(find ./exercises/practice -maxdepth 1 -mindepth 1 -type d "${name_filter[@]}" | sort)
   for exercise_dir in $all_practice_exercise_dirs; do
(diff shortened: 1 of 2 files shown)


bin/test.sh
```diff
@@ -13,6 +13,17 @@ file_ext="php"
...
+    specific_exercise=$(find ./exercises/{practice,concept} -name $1 -type d | sort)
+    for exercise_dir in $specific_exercise; do
+        test "${exercise_dir}" "example"
+        if [[ $? -ne 0 ]]; then
+          has_failures=1
+        fi
+      done
+      return $has_failures
+  fi
+
```
mk-mxp, line 26: We need to handle `practice` and `concept` differently (`example` vs. `exemplar`). So I suggest modifing the existing loops by adding the `-name` conditionally:

```suggestion
  name_filter=()
  if [ -n "$1" ]; then
    name_filter=("-name" "$1")
  fi
```

and later on:

```
  all_practice_exercise_dirs=$(find ./exercises/practice -maxdepth 1 -mindepth 1 -type d "${name_filter[@]}" | sort)
```

and

```
  all_concept_exercise_dirs=$(find ./exercises/concept -maxdepth 1 -mindepth 1 -type d "${name_filter[@]}" | sort)
```

mk-mxp (changes requested):
I like the workflow improvement! That's very welcome.

I haven't tested the suggested change, so take care. Using the `-name` filter allows wildcards like `b*`, too. Please test that this works also.

tomasnorre:
Thanks for your feedback, will have a look at in the following days.

tomasnorre:
> Using the `-name` filter allows wildcards like `b*`, too. Please test that this works also.

This works if b* is wrapped in quotes.
``` 
$ composer test:run -- "b*" 
```

tomasnorre:
I'll have to check why it's failing, will look into it.

mk-mxp:
It fails, because the shell has `set -u` (fail on use of unbound variables). Therefore `[ $# -ge 1 ] && [ -n "$1" ]` is required, as you had it.

PR #850, opened 2024-11-08
Sync book store

### LINKED ISSUE #842: Sync book-store
- [x] Run `bin/configlet sync -u -e book-store --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
  - docs: instructions unsynced: book-store
  - book-store: missing 3 test cases

tomasnorre:
Resolves #842 

There are some tests failing, with the example configuration. I had tested with my private implementation too, there some fails too. 

So I don't know if some tests aren't meant for the PHP track or what the deal is yet.

I Haven't made as detailed descriptions on the additional tests, they were more confusing than helpful to me.

### DIFF
--- a/exercises/practice/book-store/.meta/example.php
+++ b/exercises/practice/book-store/.meta/example.php
@@ -1,41 +1,42 @@
 <?php
 
-declare(strict_types=1);
-
 function total(array $items): int
 {
-    return calculate($items, 0);
-}
+    $groups = array_fill(1, 5, 0); // Groups of books
+    $sets = array_fill(1, 5, 0);   // Sets by size
 
-function calculate(array $items, int $totalCents): int
-{
-    if (empty($items)) {
-        return $totalCents;
+    // Group the basket
+    foreach ($items as $bookId) {
+        $groups[$bookId]++;
     }
 
-    $group = array_unique($items);
-    $totalMinCents = PHP_INT_MAX;
-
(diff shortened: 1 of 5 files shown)


tomasnorre:
q

tomasnorre:
Updated to match how I did it in the go-track, in that way it works.

mk-mxp:
@tomasnorre FYI: I am short of time until December. Please be patient...

exercises/practice/book-store/.meta/example.php
```diff
@@ -1,62 +1,42 @@
...
- * For more info review the Concept on strict type checking in the PHP track
- * <link>.
- *
- * To disable strict typing, comment out the directive below.
- */
-
-declare(strict_types=1);
-
-function total($items)
+function total(array $items): int
```
mk-mxp, line 3: Please adjust the students stub in `BookStore.php` to match this interface. Also add a comment into the stub, something like:

```
// Note: we expect the total in cents (1$ = 100 cents).
```

mk-mxp (changes requested):
Thanks for all the work! Just a tiny change required.

exercises/practice/book-store/.meta/example.php
tomasnorre: Will do.

mk-mxp (approved):
Thanks for your patience. And thanks for your contribution!

PR #857, opened 2024-11-12
ci: Fix strict types sniff

tomasnorre:
Resolves the problem from Updating to `"squizlabs/php_codesniffer": "3.11.0"`

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -44,7 +44,7 @@
       <property name="spacesCountAroundEqualsSign" value="0" />
     </properties>
   </rule>
-  <rule ref="src/Sniffs/ExplainStrictTypesSniff.php">
+  <rule ref="src/Sniffs/StrictTypes/ExplainStrictTypesSniff.php">
     <exclude-pattern>*/*Test\.php</exclude-pattern>
     <exclude-pattern>*/.meta/*\.php</exclude-pattern>
     <exclude-pattern>src/*</exclude-pattern>
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "php": "^8.1",
     "phpunit/phpunit": "^10.5",
     "slevomat/coding-standard": "^8.14.1",
-    "squizlabs/php_codesniffer": "^3.9"
+    "squizlabs/php_codesniffer": "^3.11.1"
   },
   "scripts": {
     "post-install-cmd": [
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "exercism/php",
   "autoload": {
-    "Exercism\\": "src",
+    "Exercism\\Sniffs\\": "src/Exercism/Sniffs",
     "Exercism\\Exercises\\": "exercises"
   },
   "config": {
--- a/src/Exercism/Sniffs/StrictTypes/ExplainStrictTypesSniff.php
+++ b/src/Exercism/Sniffs/StrictTypes/ExplainStrictTypesSniff.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Exercism\Exercism\Sniffs\StrictTypes;
+namespace Exercism\Sniffs\StrictTypes;
 
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Fixer;
--- a/src/Sniffs/StrictTypes/ExplainStrictTypesSniff.php
+++ b/src/Sniffs/StrictTypes/ExplainStrictTypesSniff.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Exercism\Sniffs;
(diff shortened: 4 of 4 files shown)


tomasnorre:
I'm a little confused that the exclude isn't respected anymore. From a quick glance all errors are from `example.php` or `*Test.php` which both should be excluded.

```
<rule ref="src/Sniffs/StrictTypes/ExplainStrictTypesSniff.php">
  <exclude-pattern>*/*Test\.php</exclude-pattern>
  <exclude-pattern>*/.meta/*\.php</exclude-pattern>
  <exclude-pattern>src/*</exclude-pattern>
  <exclude-pattern>contribution/*.php</exclude-pattern>
</rule>
```

mk-mxp:
Well, looks like the sniff name generation has gone wrong and so the excludes are not applied for our sniff. Would you mind opening an issue at PHPCodeSniffer and report the problem?

tomasnorre:
It's now working ready for review. 

I'm personally self in doubt about the `src/Exercism/` directory, if that should be called something else.

mk-mxp (changes requested):
Which version of PHP CodeSniffer is required as a minimum, so that the name change works with the `exclude-pattern`? This should be set in `composer.json` `"require-dev": { "squizlabs/php_codesniffer": "^3.11.?" }` accordingly.

phpcs.xml
```diff
@@ -44,7 +44,7 @@
       <property name="spacesCountAroundEqualsSign" value="0" />
     </properties>
   </rule>
-  <rule ref="src/Sniffs/ExplainStrictTypesSniff.php">
+  <rule ref="src/Exercism/Sniffs/StrictTypes/ExplainStrictTypesSniff.php">
```
mk-mxp, line 47: I think we should go for `src/Sniffs/StrictTypes/ExplainStrictTypesSniff.php`

That means, in `composer.json` we should have:

```json
  "autoload": {
    "Exercism\\Sniffs\\": "src/Sniffs",
    "Exercism\\Exercises\\": "exercises"
  }
```

and in `src/Sniffs/StrictTypes/ExplainStrictTypesSniff.php`

```php
namespace Exercism\Sniffs\StrictTypes;
```

tomasnorre: But then we are running into the problem again, as we don't have 4 levels deep of directories.

mk-mxp: Do we need to have a 4-level directory structure or a 4-level namespace / FQCN? I may have missed some of the discussion between you and jrnl, but it should be the namespace levels, not the file path: "The class name is already translated to a sniff name" from commit https://github.com/exercism/php/pull/857/commits/57395074618d7ac7fbc03dff48d32eeeb40d96ff of jrnl.

tomasnorre: As I understand it. 4-lever directory. But after this PR https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/676/files, not anymore if I understand it correctly. 

That is merged and released. 

Our Sniff works again, but doesn't follow conventions yet. The #857 PR should fix that though. 

As the sniffs are working again with the 3.11.1 release, I would suggest us to merge the open PRs if they are ok. Then I'll look into this afterwards again, to have it on a clean main. 

Then we can take the right approach from there.

mk-mxp: The PR from PHP_CodeSniffer you linked contains only changes handling the FQCN of sniffs. No line has anything to do with file pathes. So I still think, changing the autoload information and using the shorter directory structure is the right way to go.

tomasnorre: I'll test it out.

tomasnorre: The second I move the `Sniffs` on directory up (removing `Exercism`) out Sniffs excludes gets ignored. The sniff it's self is working, but not the `exclude-pattern`.

mk-mxp: OK, thanks for trying that. So we have to keep the file path `src/Exercism/Sniffs/StrictTypes/ExplainStrictTypesSniff.php`.

If you still happen to have time for this: We could use the `namespace Exercism\Sniffs\StrictTypes;` and adjust `composer.json` `"autoload": { "Exercism\\Sniffs\\": "src/Exercism/Sniffs" }` accordingly? That would be an improvement, too.

mk-mxp (approved):

mk-mxp (approved):

phpcs.xml
jrfnl: To be honest, for sniffs, you should generally not have a Composer `autoload` directive, but should let PHPCS handle the autoloading as the autoloading is integral to how PHPCS translates paths vs class names vs sniff codes.

If you follow the naming conventions, this will all work fine. If you don't and Composer autoloads the file, chances are that the translation between class name vs sniff code in PHPCS will not work correctly, which brings all sorts of problems.

mk-mxp: Thank your pointing at that! I didn't know this might cause problems. Followed-up in #869

PR #863, opened 2024-11-15
sync clock

### LINKED ISSUE #845: Sync clock
- [x] Run `bin/configlet sync -u -e clock --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
  - metadata: unsynced: clock
  - metadata: unsynced: clock

tomasnorre:
Resolves #845

### DIFF
--- a/exercises/practice/clock/Clock.php
+++ b/exercises/practice/clock/Clock.php
@@ -26,6 +26,17 @@ declare(strict_types=1);
 
 class Clock
 {
+    /**
+     * This class implements PHP's magic method __toString().
+     *
+     * By implementing this method, the class adheres to the `Stringable` interface.
+     * When an object of this class is used in string context (e.g., echo or string cast),
+     * this method is automatically called.
+     *
+     * More on `Stringable`: https://www.php.net/manual/en/class.stringable.php
+     *
+     * @return string The string representation of the Clock object
+     */
     public function __toString(): string
     {
         throw new \BadMethodCallException("Implement the __toString function");
--- a/exercises/practice/clock/ClockTest.php
+++ b/exercises/practice/clock/ClockTest.php
@@ -1,27 +1,5 @@
 <?php
 
-/*
- * By adding type hints and enabling strict type checking, code can become
- * easier to read, self-documenting and reduce the number of potential bugs.
- * By default, type declarations are non-strict, which means they will attempt
- * to change the original type to match the type specified by the
- * type-declaration.
- *
- * In other words, if you pass a string to a function requiring a float,
- * it will attempt to convert the string value to a float.
- *
- * To enable strict mode, a single declare directive must be placed at the top
- * of the file.
- * This means that the strictness of typing is configured on a per-file basis.
- * This directive not only affects the type declarations of parameters, but also
- * a function's return type.
- *
- * For more info review the Concept on strict type checking in the PHP track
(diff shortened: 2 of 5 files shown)


exercises/practice/clock/ClockTest.php
```diff
@@ -31,204 +9,527 @@ public static function setUpBeforeClass(): void
...
+    /**
+     * uuid a577bacc-106b-496e-9792-b3083ea8705e
+     * @testdox on the hour
+     */
     public function testOnTheHour(): void
     {
         $clock = new Clock(8);
 
-        $this->assertEquals('08:00', $clock->__toString());
+        $this->assertEquals('08:00', $clock);
```
mk-mxp, line 20: This is one of the magic tricks in PHP, but we currently have no concept (concept exercise) for students to learn this. So we should either explain the magic in the stub (for now) or explicitly call the method as done before:

```suggestion
        $this->assertEquals('08:00', $clock->__toString());
```

or

```suggestion
        $this->assertEquals('08:00', (string)$clock);
        // plus use `implements Stringable` in the students stub
        // plus a comment in the stub pointing to PHP docs on interface `Stringable`
```

@homersimpsons What do you think about this?

mk-mxp (changes requested):
Also we should have a comment in the students stub that missing methods need to be added.

exercises/practice/clock/ClockTest.php
```diff
@@ -31,204 +9,527 @@ public static function setUpBeforeClass(): void
...
+    {
+        $clock = new Clock(2, 20);
+        $clock = $clock->sub(3000);
+        $this->assertEquals("00:20", $clock);
     }
 
-    public function testEquivalentClocks(): void
+    /**
+     * uuid f2fdad51-499f-4c9b-a791-b28c9282e311
+     * @testdox clocks with same time
```
mk-mxp, line 394: And the last outer `cases` change:

```suggestion
     * @testdox Compare two clocks for equality - clocks with same time
```

mk-mxp (changes requested):
These changes are not connected to the discussion about the magic method usage.

Please add a comment to the students stub, that all missing methods need to be added by the student.

exercises/practice/clock/ClockTest.php
homersimpsons: If we want to show magic methods with this exercise, I think I would prefer the explicit `(string)` cast, it has the advantages to clearly show that this method is magic.

But I do not know if we want to show those here.

I believe we could have a concept to show `__toString`, `__set`, `__get` which I think are rather "common". `__construct` is a magic method but I believe it is already being taught.

exercises/practice/clock/Clock.php
```diff
@@ -26,6 +26,17 @@
...
+     * This class implements PHP's magic method __toString().
+     *
+     * By implementing this method, the class adheres to the `Stringable` interface.
+     * When an object of this class is used in string context (e.g., echo or string cast),
+     * this method is automatically called.
+     *
+     * More on `Stringable`: https://www.php.net/manual/en/class.stringable.php
+     *
+     * @return string The string representation of the Clock object
+     */
```
mk-mxp, line 39: Great!

mk-mxp (approved):
Thanks for the contribution.

PR #867, opened 2024-11-19
sync darts

### LINKED ISSUE #853: Sync darts
- [x] Run `bin/configlet sync -u -e darts --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
  - docs: instructions unsynced: darts
  - metadata: unsynced: darts
  - darts: missing 13 test cases

tomasnorre:
Resolves #853

### DIFF
--- a/exercises/practice/darts/.meta/example.php
+++ b/exercises/practice/darts/.meta/example.php
@@ -1,27 +1,5 @@
 <?php
 
-/*
- * By adding type hints and enabling strict type checking, code can become
- * easier to read, self-documenting and reduce the number of potential bugs.
- * By default, type declarations are non-strict, which means they will attempt
- * to change the original type to match the type specified by the
- * type-declaration.
- *
- * In other words, if you pass a string to a function requiring a float,
- * it will attempt to convert the string value to a float.
- *
- * To enable strict mode, a single declare directive must be placed at the top
- * of the file.
- * This means that the strictness of typing is configured on a per-file basis.
- * This directive not only affects the type declarations of parameters, but also
- * a function's return type.
- *
- * For more info review the Concept on strict type checking in the PHP track
- * <link>.
- *
- * To disable strict typing, comment out the directive below.
- */
-
 declare(strict_types=1);
 
 class Darts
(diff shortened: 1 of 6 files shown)


mk-mxp:
Discussion about a re-make in #853

exercises/practice/darts/.meta/example.php
```diff
@@ -1,52 +1,20 @@
...
- * For more info review the Concept on strict type checking in the PHP track
- * <link>.
- *
- * To disable strict typing, comment out the directive below.
- */
-
 declare(strict_types=1);
 
-class Darts
+function calculateScore(float $xAxis, float $yAxis): int
```
mk-mxp, line 5: Canonical data uses `"property": "score",`, so we should use `score` as the functions name:

```suggestion
function score(float $xAxis, float $yAxis): int
```

exercises/practice/darts/Darts.php
```diff
@@ -24,12 +24,7 @@
 
 declare(strict_types=1);
 
-class Darts
+function __construct(float $xAxis, float $yAxis)
```
mk-mxp, line 27:
```suggestion
function score(float $xAxis, float $yAxis): int
```

exercises/practice/darts/Darts.php
```diff
@@ -24,12 +24,7 @@
...
 {
-    public int $score = 0;
-
-    public function __construct(float $xAxis, float $yAxis)
-    {
-        throw new BadFunctionCallException("Please implement the Darts class!");
-    }
+    throw new BadFunctionCallException("Please implement the Darts class!");
```
mk-mxp, line 29:
```suggestion
    throw new BadFunctionCallException("Please implement the __FUNCTION__ function!");
```

mk-mxp (changes requested):
Thanks for taking care of this!

exercises/practice/darts/.meta/example.php
tomasnorre: My bad, will correct that later.

mk-mxp (approved):
Thanks a lot!

PR #870, opened 2024-11-21
Remove autoloading for PHP_CodeSniffer Sniff

### LINKED ISSUE #869: Remove autoloading for PHP_CodeSniffer sniff
As recommended by the maintainer of PHP_CodeSniffer, we should not add the sniff to `composer` autoloading.

> To be honest, for sniffs, you should generally not have a Composer `autoload` directive, but should let PHPCS handle the autoloading as the autoloading is integral to how PHPCS translates paths vs class names vs sniff codes.
> 
> If you follow the naming conventions, this will all work fine. If you don't and Composer autoloads the file, chances are that the translation between class name vs sniff code in PHPCS will not work correctly, which brings all sorts of problems.

_Originally posted by @jrfnl in https://github.com/exercism/php/pull/857#discussion_r1848762676_

tomasnorre:
Resolves #869 

To be honest I'm a little in doubt if this is all that is meant with the issue.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "exercism/php",
   "autoload": {
-    "Exercism\\": "src/Exercism",
+    "Exercism\\": "src",
     "Exercism\\Exercises\\": "exercises"
   },
   "config": {
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "exercism/php",
   "autoload": {
-    "Exercism\\Sniffs\\": "src/Exercism/Sniffs",
+    "Exercism\\": "src/Exercism",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "exercism/php",
   "autoload": {
-    "Exercism\\Sniffs\\": "src/Exercism/Sniffs",
+    "Exercism\\": "src",
```
mk-mxp, line 4: Remove that entry completely, so composer does not handle `Exercism\*` or `Exercism\Sniffs\*`:

```suggestion
```

The other entry for exercises does no harm, but currently also does not do anything used by the track.

mk-mxp (commented):
Thanks for looking into it. Small steps, that help.

PR #881, opened 2025-03-01
Auto-sync instructions

mk-mxp:
Auto-sync with problem specifications.

For `sieve` there was [a lengthy discussion on the forum](https://forum.exercism.org/t/updating-sieve-exercise-instructions/15703) about the "using the proper algorithm" note, namely about the use of division or modulo for the prime detection. But I don't think we should change our `example.php` to match that - it uses modulo to detect multiples instead of looping over multiples of a prime to remove them.

### DIFF
--- a/exercises/practice/affine-cipher/.docs/instructions.md
+++ b/exercises/practice/affine-cipher/.docs/instructions.md
@@ -20,7 +20,7 @@ Where:
 
 - `i` is the letter's index from `0` to the length of the alphabet - 1.
 - `m` is the length of the alphabet.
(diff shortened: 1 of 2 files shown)


homersimpsons:
(I will review this tonight or tomorrow)

homersimpsons (approved):

homersimpsons:
> But I don't think we should change our example.php to match that - it uses modulo to detect multiples instead of looping over multiples of a prime to remove them.

This file is in fact used to prove that the exercise is solvable. I think some student may read it though. I'm fine keeping it like this.

PR #882, opened 2025-03-05
update php/docs/intallation.md

### LINKED ISSUE #864: Discussion: Improve PHP track docs for PHPUnit / composer
I think [our track docs](https://github.com/exercism/php/blob/main/docs/INSTALLATION.md) make things too complicated for new students.

Right after the required installation of PHP we have a short paragraph about [installing `composer`](https://github.com/exercism/php/blob/main/docs/INSTALLATION.md#install-composer). For PHPUnit we then describe `composer` as the main means of installation. Using the PHPUnit PHAR is last and "optional". But composer is **not** required to solve Exercism exercises.

The easiest way to use PHPUnit for Exercism is downloading the PHAR. So we should provide tutorial-style instructions on that first. Also we should check, that [`exercism test`](https://github.com/exercism/cli/blob/7580d02cbb5b0904d0e72e9e9782e61c734265a3/workspace/test_configurations.go#L197) runs the PHAR when using these instructions.

In a more optional looking paragraph we may then describe and explain `composer` and PHPUnit installation using it. There is a nice short [description of composer and installation](https://github.com/adambard/learnxinyminutes-docs/blob/master/php-composer.html.markdown) we may use to improve the `composer` introduction and installation. We should also mention that `composer` is wide-spread in the PHP ecosystem.

A-O-Emmanuel:
Edit: Closes #864

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -101,7 +101,7 @@ chmod +x phpunit.phar
 ```
 Now you can run the PHAR.
 
-You can also follow the official  [Installing PHPUnit instructions](https://docs.phpunit.de/en/10.5/installation.html#installing-phpunit) to Install PHPUnit via a PHP Archive (PHAR) 
+You can also follow the official [Installing PHPUnit instructions](https://docs.phpunit.de/en/10.5/installation.html#installing-phpunit) to install PHPUnit.
 
 ### Via Composer
 
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -99,7 +99,26 @@ Then make the downloaded file executable:
 ```shell
 chmod +x phpunit.phar
 ```
-Now you can run the PHAR.
+
+And move it to the same location as you moved the `exercism` CLI, naming the command `phpunit`:
+
+```shell
+mv phpunit.phar ~/bin/phpunit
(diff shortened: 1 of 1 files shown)


mk-mxp:
Thanks for your contribution! I'm in the process of testing the workflow in a new virtual machine. As I have very limited time at the moment, this may take some time. Hopefully no more than a day or two...

mk-mxp (changes requested):
We need to make `phpunit` available as a global command like the `exercism` CLI to enable the testing workflow shown in the [PHP track testing docs](https://exercism.org/docs/tracks/php/tests).

Also, Markdown requires empty lines before / after code blocks and headings. There is a load of [Markdown docs for Exercism](https://exercism.org/docs/building/markdown), but you don't have to read the docs for this PR.

A-O-Emmanuel:
@mk-mxp I committed the suggestiong you made but i don't quite understand when you say "We need to make phpunit available as a global command like the exercism CLI to enable the testing workflow shown in the [PHP track testing docs](https://exercism.org/docs/tracks/php/tests)." Looks this is already part of the suggestions you made?

mk-mxp (approved):

A-O-Emmanuel:
Thank you

PR #887, opened 2025-04-14
Sync largest-series-product tests

mk-mxp:
This does not change anything but UUIDs, as we do not check for Exception messages here.


tomasnorre (approved):

homersimpsons (approved):
Approved, don't forget to add `[no important files changed]` during merge

PR #894, opened 2025-05-31
docs: change use to used

Edwin-devc:


mk-mxp (changes requested):
Hi @Edwin-devc, thanks for bringing this up!

The file you edited is derived from [`concepts/booleans/introduction.md`](https://github.com/exercism/php/blob/main/concepts/booleans/introduction.md), where the same fix needs to be made. The failing CI task checks for the two places being identical.

Looking forward to merge this after you fixed the second file, also.

Thanks for contributing!

Edwin-devc:
Hi @mk-mxp, thanks for the heads-up!

I've applied the same fix to concepts/booleans/introduction.md as well. The two files should now be identical, and the CI check should pass.

Let me know if anything else is needed.

Thanks again!

PR #901, opened 2025-07-05
Auto-sync exercises

mk-mxp:
* Add parameters to use for sync'ing the tests in a second run
* Cleanup auto-sync list
* Bugfix in docs of `luhn`
* Re-make of `protein-translation` instructions, see [this](https://forum.exercism.org/t/tweak-language-and-formatting-in-protein-translation-description-md/17005) and [this](https://forum.exercism.org/t/cutting-the-mention-of-how-many-codons-there-are/17310) forum threads

### DIFF
--- a/bin/auto-sync.txt
+++ b/bin/auto-sync.txt
@@ -48,9 +48,6 @@ phone-number
 pig-latin
 protein-translation
 raindrops
-resistor-color
-resistor-color-duo
-resistor-color-trio
 reverse-string
 rna-transcription
 robot-simulator


bin/auto-sync.sh
```diff
@@ -5,6 +5,7 @@ if [[ "$#" != 0 ]]; then
     forwardedParameters=( "$@" )
 else
     forwardedParameters=( -o -u -y --docs --filepaths --metadata )
+    # forwardedParameters=( -o -u -y --tests include )
```
homersimpsons, line 8: Do we want to commit this?

```suggestion
```

homersimpsons (approved):

bin/auto-sync.sh
mk-mxp: I've been looking these up every month again from `configlet` help and want to have them at hand now.

Issue #884, opened 2025-04-14
Coordination: Upgrade to PHP 8.2 - 8.4 and PHPUnit 11


mk-mxp:
This is what I think must be done:

- [x] #902 to the CI of the track and allow it to fail
- [x] #904 Use PHPUnit attributes, not annotations (annotations are deprecated, but still available)
  - [x] Update test generator to produce attributes not annotations
  - [x] Convert tests from annotations to attributes. Use `rector/rector` for that
- [x] Fix the problems that occur with PHP 8.4 and PHPUnit 11
  - [x] Measure performance when using PHP 8.4 (look at `alphametics`!)
  - [x] Check the CI test to catch tests that would fail due to the runtime limit set by Exercism infrastructure. Uses `timeout 54s <command>` as a process based limit per exercise (3x 18 seconds, see [forum for calculation details](https://forum.exercism.org/t/test-tracks-for-the-20-seconds-limit-on-test-runners/10536/8))
- [x] Prepare the [PHP test-runner](https://github.com/exercism/php-test-runner) to upgrade to PHP 8.4 / PHPUnit11: https://github.com/exercism/php-test-runner/issues/150
- [x] Prepare the [PHP representer](https://github.com/exercism/php-representer) to upgrade to PHP 8.4: https://github.com/exercism/php-representer/issues/297
- [x] #903  as the production versions
  - [x] https://github.com/exercism/php-test-runner/pull/157 - Merge test runner upgrade PR
  - [x] Update student facing documentation
  - [x] Remove PHP 8.1 / PHPUnit 10 from CI, add PHP8.2/8.3 to PHPUnit11 workflow
  - [x] Revert performance workaround in "robot-name" exercise (requires PHPUnit >= 11)

In the whole process, try to avoid re-running the representer and not to trigger re-testing all community solutions. These are very costly to Exercism.

FYI: @tomasnorre @homersimpsons

mk-mxp:
@homersimpsons PHPUnit 12 does not support PHP8.2, which is the minimum we want to support. Maybe we can discuss later, if we step away from the "3 latest minor PHP versions" principle. Right now, PHPUnit 11 is the only PHPUnit version suitable for the range of PHP 8.2 - 8.4.

tomasnorre:
I didn't pay attention to that requirement.

mk-mxp:
Now it's time for `rector`. Has anyone time to apply it? @tomasnorre @homersimpsons

PR #902, opened 2025-07-05
Add PHP 8.4 and PHPUnit 11.5 testing

mk-mxp:
This is part of #884 

* Add workflow file for PHPUnit 11 using PHP8.4 on Ubuntu 24.04
* Allow the job to fail in CI (`continue-on-error: true`)

Running it emits the expected PHPUnit deprecations for annotations (e.g. for Windowing System: `PHPUnit Deprecations: 9.`). This is already in #884

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ The `lint:check` is included in `composer ci` to run the CI checks locally.
 - Follow the [PSR-12] coding style (Exercisms PHP track uses a slightly [modified][local-file-phpcs-config] version of [PSR-12]).
 - Run `composer ci` to run CI checks locally before pushing.
 - CI is run on all pull requests, it must pass the required checks for merge.
-- CI is running all tests on PHP 8.0 to PHP 8.2 for Linux, Windows and MacOS.
+- CI is running all tests on PHP 8.1 to PHP 8.4 for Linux, Windows and MacOS.
 
 ## Generating new practice exercises
 


.github/workflows/exercise-tests-phpunit-11.yml
```diff
@@ -0,0 +1,48 @@
...
+  test:
+    name: PHP ${{ matrix.php-version }} - ${{ matrix.os }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # php-version: [8.2, 8.3, 8.4]
+        # os: [ubuntu-24.04, windows-2022, macOS-14]
+        php-version: [8.4]
+        os: [ubuntu-24.04]
```
homersimpsons, line 20: Should this be updated?

.github/workflows/exercise-tests-phpunit-11.yml
```diff
@@ -0,0 +1,48 @@
...
+      - uses: shivammathur/setup-php@0f7f1d08e3e32076e51cae65eb0b0c871405b16e
+        with:
+          php-version: ${{ matrix.php-version }}
+          extensions: gmp
+
+      - name: Install dependencies
+        shell: bash
+        run: |
+          curl -Lo ./bin/phpunit-11.phar https://phar.phpunit.de/phpunit-11.phar
+          chmod +x bin/phpunit-11.phar
```
homersimpsons, line 40: Should we use composer here?

.github/workflows/exercise-tests-phpunit-11.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Test exercises
+        continue-on-error: true
```
homersimpsons, line 43: Is this expected?

.github/workflows/exercise-tests-phpunit-11.yml
mk-mxp: I currently only want one additional CI job, not 9. When "going into production" all of these versions and OSses should be activated.

.github/workflows/exercise-tests-phpunit-11.yml
mk-mxp: I thought of that, but had no time to fiddle with the details. So I used the copy of the previous version for the ease of the task.

.github/workflows/exercise-tests-phpunit-11.yml
mk-mxp: CI actually seems to pass, but I do not want any upcoming things to hang because of the currently experimental support of PHP 8.4 / PHPUnit 11.5.

E.g. in the test runner, I currently have problems with Gigaseconds. The dataProviders there cause errors, that stop PHPUnit 11 from working but are fine with PHPUnit 10. Such should not stop the production CI from passing.

PR #903, opened 2025-07-07
Draft: Update for PHP 8.4 / PHPUnit 11.5

mk-mxp:
As mentioned in https://github.com/exercism/php-test-runner/pull/157, we can directly upgrade the track to the higher versions and do the annotation -> attribute conversion later.

Do not merge!


docs/INSTALLATION.md
```diff
@@ -77,12 +77,12 @@ You will also need [Docker](https://docs.docker.com/engine/install/).
...
-## Install Composer
-
-Install [Composer](https://getcomposer.org) following your devices OS [installation instructions](https://getcomposer.org/doc/00-intro.md). We recommend installing it globally for ease of use.
-
 ## Install PHPUnit
 
+~~~~exercism/note
+PHPUnit 12 is not supported, yet!
+~~~~
+
```
homersimpsons, line 85: Do we want to mention it as we provide the installation steps below? Or should we just mention that we are using PHPUnit 11 to run the tests. I believe most of the persons seing this won't know about phpunit 12.

homersimpsons (approved):
Perfect, thanks! I just have a doc comment but I'm completely fine merging this as-is.

docs/INSTALLATION.md
mk-mxp: I don't know if this note box really helps, but we do actually get [help requests because of wrong versions](https://forum.exercism.org/t/solved-same-error-argumentcounterror-on-2-php-exercises/16475). That means, people don't follow our instructions but use the official docs. And hopefully now read beforehand which version they cannot use, before they switch to the official docs.

PR #904, opened 2025-07-08
:recycle: Use PHPUnit's attributes in place of annotations (#884)

homersimpsons:
1. Used rector with the following configuration:
     ```php
     <?php
    declare(strict_types=1);

    use Rector\Config\RectorConfig;
    use Rector\PHPUnit\Set\PHPUnitSetList;

    return RectorConfig::configure()
        ->withPaths([__DIR__ . '/exercises'])
        ->withSets([PHPUnitSetList::ANNOTATIONS_TO_ATTRIBUTES, PHPUnitSetList::PHPUNIT_110])
	->withImportNames(importShortClasses: false, removeUnusedImports: true);
    ```
2. Manually fix testdox that contains `::`:
    ```diff
    -#[TestDox(specify a string type for Address::$street)]
    +#[TestDox('specify a string type for Address::$street')]
    ```
4. Updated `TestGenerator.php` to use attribute
5. Updated `bootstrap.sh` to use attribute

### DIFF
--- a/contribution/generator/src/TestGenerator.php
+++ b/contribution/generator/src/TestGenerator.php
@@ -53,7 +53,11 @@ class TestGenerator
             $method = $this->builderFactory->method($methodName)
                 ->makePublic()
                 ->setReturnType('void')
-                ->setDocComment("/**\n * uuid: {$case->uuid}\n * @testdox {$description}\n */")
+                ->setDocComment("/** uuid: {$case->uuid} */")
+                ->addAttribute(new Node\Attribute(
+                    new Node\Name(\PHPUnit\Framework\Attributes\TestDox::class),
+                    [new Node\Arg(new Node\Scalar\String_($description))],
+                ))
                 ->addStmt(
                     $this->builderFactory->funcCall(
                         '$this->markTestIncomplete',
--- a/exercises/concept/bootstrap.sh
+++ b/exercises/concept/bootstrap.sh
@@ -70,17 +70,18 @@ PHP_STUB
 cat <<- PHP_STUB >> "${base_dir}/${test_file}"
 <?php
 
-class ${new_classname}Test extends PHPUnit\Framework\TestCase
+use PHPUnit\Framework\TestCase;
+use PHPUnit\Framework\Attributes\TestDox;
+
+class ${new_classname}Test extends TestCase
 {
     public static function setUpBeforeClass(): void
     {
         require_once '${new_classname}.php';
     }
 
-    /**
-     * @testdox some test stub
-     * @task_id 1
-     */
+    /** @task_id 1 */
+    #[TestDox('some test stub')]
     public function testStub()
     {
         \$class = new ${new_classname}();
--- a/exercises/practice/darts/Darts.php
+++ b/exercises/practice/darts/Darts.php
@@ -26,5 +26,5 @@ declare(strict_types=1);
 
 function score(float $xAxis, float $yAxis): int
 {
-    throw new BadFunctionCallException("Please implement the __FUNCTION__ function!");
+    throw new \BadFunctionCallException("Please implement the __FUNCTION__ function!");
 }
--- a/exercises/practice/isogram/IsogramTest.php
+++ b/exercises/practice/isogram/IsogramTest.php
@@ -2,7 +2,10 @@
 
 declare(strict_types=1);
 
-class IsogramTest extends PHPUnit\Framework\TestCase
+use PHPUnit\Framework\TestCase;
+use PHPUnit\Framework\Attributes\TestDox;
+
+class IsogramTest extends TestCase
 {
     public static function setUpBeforeClass(): void
     {
@@ -11,8 +14,8 @@ class IsogramTest extends PHPUnit\Framework\TestCase
 
     /**
      * uuid a0e97d2d-669e-47c7-8134-518a1e2c4555
-     * @testdox empty string
      */
+    #[TestDox('empty string')]
     public function testEmptyString(): void
     {
         $this->assertTrue(isIsogram(''));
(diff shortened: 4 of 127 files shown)


exercises/practice/darts/Darts.php
```diff
@@ -26,5 +26,5 @@
 
 function score(float $xAxis, float $yAxis): int
 {
-    throw new BadFunctionCallException("Please implement the __FUNCTION__ function!");
+    throw new \BadFunctionCallException("Please implement the __FUNCTION__ function!");
```
homersimpsons, line 29: Somehow this one did not had a `\` while all others have.

mk-mxp:
@homersimpsons So quick and so many changes :) I hope to get to it on the weekend.

homersimpsons:
> @homersimpsons So quick and so many changes :) I hope to get to it on the weekend.

Yes, it took me 30 minutes to 1 hour to check the whole diff. Most of the files had the same changes, also the diff count per file also gives confidence at a quick glance.

contribution/generator/src/TestGenerator.php
```diff
@@ -53,7 +53,11 @@ public function createTestsFor(
             $method = $this->builderFactory->method($methodName)
                 ->makePublic()
                 ->setReturnType('void')
-                ->setDocComment("/**\n * uuid: {$case->uuid}\n * @testdox {$description}\n */")
+                ->setDocComment("/**\n * uuid: {$case->uuid}\n */")
```
mk-mxp, line 56: My preference would be:

```suggestion
                ->setDocComment("/** uuid: {$case->uuid} */")
```

contribution/generator/src/TestGenerator.php
```diff
@@ -53,7 +53,11 @@ public function createTestsFor(
...
+                ->addAttribute(new Node\Attribute(
+                    new Node\Name(\PHPUnit\Framework\Attributes\TestDox::class),
+                    [new Node\Arg(new Node\Scalar\String_($description))],
```
mk-mxp, line 59:
```suggestion
                    [new Node\Arg(new Node\Scalar\String_(ucfirst($description)))],
```

In case we assume PHP8.4 for using the generator: `mb_ucfirst()`.

exercises/concept/bootstrap.sh
```diff
@@ -70,17 +70,20 @@ PHP_STUB
...
 {
     public static function setUpBeforeClass(): void
     {
         require_once '${new_classname}.php';
     }
 
     /**
-     * @testdox some test stub
      * @task_id 1
      */
```
mk-mxp, line 85: My preference is `/** @task_id 1 */`, if you agree with me.

exercises/practice/darts/Darts.php
mk-mxp: Thanks for spotting this!

exercises/practice/isogram/IsogramTest.php
```diff
@@ -11,127 +14,126 @@ public static function setUpBeforeClass(): void
...
     }
 
     /**
      * uuid 0d0b8644-0a1e-4a31-a432-2b3ee270d847
-     * @testdox word with duplicated character and with two hyphens
-     *
-     * This test aims to catch buggy implementations that check for duplicate spaces or hyphens.
      */
+    #[TestDox('word with duplicated character and with two hyphens
+This test aims to catch buggy implementations that check for duplicate spaces or hyphens.')]
```
mk-mxp, line 136:
```suggestion
    #[TestDox('word with duplicated character and with two hyphens')]
```

I think, the rest of the comment wasn't meant to be student facing.

exercises/practice/reverse-string/ReverseStringTest.php
```diff
@@ -11,53 +14,51 @@ public static function setUpBeforeClass(): void
...
     {
         $this->assertEquals("", reverseString(""));
     }
 
-    /**
-     * @testdox a word
-     * uudi 01ebf55b-bebb-414e-9dec-06f7bb0bee3c
-     */
+    #[TestDox('a word
+uudi 01ebf55b-bebb-414e-9dec-06f7bb0bee3c')]
```
mk-mxp, line 25: Ups:

```suggestion
    /** uuid 01ebf55b-bebb-414e-9dec-06f7bb0bee3c */
    #[TestDox('a word')]
```

mk-mxp (changes requested):
Only a few problems, automation is great!

One thing I am unsure about: Why do we have `\` on students stubs exceptions, but not on those in the tests? Technically `\` is not required, as there are no namespaces set. Should we remove them?

contribution/generator/src/TestGenerator.php
homersimpsons: Applied. I find this debatable, personally I usually prefer to avoid one-line doc comment for functions and classes.

Note that I didn't update the files to reflect this change in all the generated files, do you think that I should do so?

contribution/generator/src/TestGenerator.php
homersimpsons: This is already done on line 48: https://github.com/homersimpsons/exercism-php/blob/33ebb3bc26f00df67921e5708d519a11752a2ca3/contribution/generator/src/TestGenerator.php#L48

exercises/concept/bootstrap.sh
homersimpsons: Same topic as uuid, should I update it globally?

exercises/practice/reverse-string/ReverseStringTest.php
homersimpsons: Applied (with inlined uuid)

exercises/concept/bootstrap.sh
mk-mxp: Dito. Not required.

mk-mxp (approved):
Thanks a lot for contributing this!

PR #908, opened 2025-08-10
Auto-sync tests also by default

mk-mxp:
The amount of test changes in the past monthes has been very low, we should be able to handle them every month.


homersimpsons (approved):
I'm not familiar with this but it looks fine. Thanks.

PR #915, opened 2025-08-14
Linked list

A-O-Emmanuel:
Synced `linked-list` exercise using `configlet`.

Removed `strict_types` comments.

Added and synced test metadata (`uuid` / `#TestDox`) to tests.

Added `linked-list` to `bin/auto-sync.txt` (alphabetically sorted).

### DIFF
--- a/exercises/practice/linked-list/.meta/config.json
+++ b/exercises/practice/linked-list/.meta/config.json
@@ -2,7 +2,7 @@
   "authors": [
     "MichaelBunker"
   ],
-  "contributors":[
+  "contributors": [
     "A-O-Emmanuel"
   ],
   "files": {


mk-mxp (changes requested):
Thanks for taking this forward! Now comes _the hard part_ :smile: - You will learn something about the Exercism practice exercise ecosystem...

The file `.meta/tests.toml` lists all the tests, that are in the [canonical data JSON](https://github.com/exercism/problem-specifications/blob/main/exercises/linked-list/canonical-data.json). That means, we **don't have** many of those tests, yet. And those we have do not use the values from canonical data.

Please sync the existing tests: Use the UUID to find the test data used in canonical data JSON and adjust our tests. Run the tests using `composer test:run -- linked-list` to see, if they still pass.

Then add the missing tests, one after the other. Name the test methods accordingly (`test` + camelCasedDescription). When you run the tests, and the new test fails, comment out the test method and mention that test in this Pull Request. We need then need to discuss if breaking existing things is OK.

If you have any questions, please ask them here. Thanks for helping us!

A-O-Emmanuel:
It's definitely the hard part, lol, well, I'm happy to help and yes, I'll ask if I have any questions, thank you.

A-O-Emmanuel:
HI @mk-mxp I've added the missing tests, synced the existing test values with the values in the canonical data and also commented out the tests that failed because the count and delete functions where not implemented. These are the tests descriptions that failed: 

count an empty list

count a list with items

count is correct after mutation

popping to empty doesn't break the list

shifting to empty doesn't break the list

deletes the only element

deletes the element with the specified value from the list

deletes the element with the specified value from the list, re-assigns tail

deletes the element with the specified value from the list, re-assigns head

deletes the first of two elements

deletes the second of two elements

delete does not modify the list if the element is not found

deletes only the first occurrence

mk-mxp:
> commented out the tests that failed because the count and delete functions where not implemented

That looks like it is most of the exercise! The [Track Build Status](https://exercism.org/tracks/php/build) page says in "Practice Exercises" -> "...active practice exercises" that there is only 102 solutions to this exercise, yet. So I'd say, let's break the existing solutions and add all the tests.

@A-O-Emmanuel Please uncomment the tests and adjust `.meta/example.php` to pass them. If you want to, you can remove the current example and start from the students stub, just as you would do when solving the exercise :smiley: Or, if you think that is beyond what you can do, ask here for someone else to make the tests pass.

I hope to have time for a good review on the weekend.

A-O-Emmanuel:
Hi @mk-mxp I've adjusted example.php to pass the tests, please take a look and tell me if there's more to be done. Thank you

mk-mxp (approved):
Thank you very much! Many people can now learn more about linked lists than before.

A-O-Emmanuel:
My pleasure!

PR #917, opened 2025-08-25
Fix description for random key in SimpleCipher

IsaacUtah1379:
The tests in SimpleCipher expect the key to only contain lowercase letters, but the instructions for generating a random key say to use alphanumeric characters.


IsaacUtah1379:
I'd be happy to! Is there anything special I need to do, or should I just copy the description from problem-specifications over to the instructions.md file?

mk-mxp:
@IsaacUtah1379 please simply copy the instructions.

PR #919, opened 2025-09-02
Add SymfonyCasts to LEARNING.md

tomasnorre:


mk-mxp:
While in general it is good to know about Symfony when going into PHP for development, it is not in the scope of Exercism to teach about frameworks or 3rd party libraries. So I think we should not add this.

@homersimpsons What do you think?

homersimpsons:
I don't really have a strong opinin on this. To me we should focus on PHP the language itself. But we can certainly add framework ressources in a "going further" section.

That said, it does not seem that other tracks push their frameworks, for example Ruby which has the popular Ruby on Rails does not have it in its list https://exercism.org/docs/tracks/ruby/learning (same for Java/Spark, Dart/Flutter, JavaScript/any).

I do not know if students often go to those pages but I think we can improve our presentation here, having:
1. Official links (php doc)
2. Unofficial references / Tutorials (text / video) with short descriptions (for instance we could add https://learnxinyminutes.com/php/)
3. QA resources (stack overflow)
4. Going further resources: things that are linked to the language such as frameworks, orm, observations

mk-mxp:
@tomasnorre Thanks for accepting!

PR #934, opened 2025-11-11
Add test for invalid board with wrong move count for winner

Caminaur:
This PR adds a missing test for the Tic-Tac-Toe exercise, covering the case
where X wins but both players have the same move count, which represents an
impossible game state.


mk-mxp:
@Caminaur Thanks for bringing this up! This exercise takes its test cases from the [common problem specification repository](https://github.com/exercism/problem-specifications/blob/main/exercises/state-of-tic-tac-toe/canonical-data.json). Please start a [forum thread](https://forum.exercism.org/), so all maintainers can vote for or against this new test.

PR #935, opened 2025-11-15
Fix mismatched hints for Pizza Pi exercise

mikefizzled:
- Align hints with the current instructions
- Correct rounding and pi() references
- Note that explicit ceil()/floor() references could be added later if desired

### DIFF
--- a/exercises/concept/pizza-pi/.docs/hints.md
+++ b/exercises/concept/pizza-pi/.docs/hints.md
@@ -25,5 +25,5 @@
 - Use the [modulo operator][arithmetic-operations] (`%`) to get the remainder after division.
 
 [arithmetic-operations]: https://www.php.net/manual/en/language.operators.arithmetic.php
-[pi-function]: https://www.php.net/manual/en/function.pi.php
+[pi-constant]: https://www.php.net/manual/en/function.pi.php
 [numeric-functions]: https://www.php.net/manual/en/ref.math.php


exercises/concept/pizza-pi/.docs/hints.md
```diff
@@ -6,23 +6,22 @@
...
 ## 2. A Splash of Sauce
 
-- When rounding up, PHP provides a [numeric function][numeric-functions].
+- To ensure a whole number of cans, you'll need to round up. PHP provides [numeric functions][numeric-functions] that can help.
 
 ## 3. Some Cheese, Please
 
-- For cubing a number, there is a [built-in exponentiation function][arithmetic-operations].
+- You can cube a number using the [exponentiation operator][arithmetic-operations].
+- You’ll need to round the result down; PHP has a [numeric function][numeric-functions] for that.
```
mk-mxp:
```suggestion
- To ensure a whole number of pizzas, you’ll need to round down.
- PHP provides [numeric functions][numeric-functions] that can help.
```

exercises/concept/pizza-pi/.docs/hints.md
```diff
@@ -6,23 +6,22 @@
...
+- PHP also provides a [built-in function][pi-function] for pi (π)
```
mk-mxp: Usually the constant is prefered, as it avoids the function call overhead.

```suggestion
- PHP has a [built-in constant][pi-constant] for pi (π).
```

mk-mxp (changes requested):
Thanks for taking care of the hints! They really needed some corrections.

At Exercism, we try to give short hints per line, so one can stop reading as soon as one is unstuck. Also we want to speak to the person reading, so we try to avoid "this task" as the subject.

Anyway, my comments are suggestions. Feel free to pick those you like and reject the others.

mk-mxp (approved):
Thanks for your contribution! The changes will be live in a few minutes...

PR #939, opened 2025-11-24
Make instructions consistent with (y, x) and (height, width) ordering

mikefizzled:
To minimise confusion within the exercise, the instructions now consistently refer to the parameters in the order they are required. While the ordering is a little strange, it maintains the backwards compatibility of completed exercises.

All instructions now follow:

(y, x)
(height, width)


mk-mxp:
Thank you for fixing this!

PR #948, opened 2026-03-15
fix: Refactor exception message to display function name properly

jawira:
I refactored exception message since constants are not interpolated.


mk-mxp (approved):
Thanks for fixing this!

PR #949, opened 2026-03-28
Sync 2026-04

mk-mxp:
* Mostly documentation links
* New test cases for `connect`, which might break some existing solutions

We should re-test all existing solutions.


homersimpsons (approved):
Thanks for the sync !

PR #951, opened 2026-04-15
bin/verify-exercises-in-docker

keiravillekode:
This script is useful for testing against a test runner image.

We remove the affine-cipher solution's dependence on GMP.


mk-mxp (approved):
Thanks for this contribution!

PR #972, opened 2026-05-24
Fix typo in language-list (phpunit test)  [no important files changed]

resu-xuniL:
This pull request fix a typo in LanguageListTest.php
[no important files changed]

### DIFF
--- a/exercises/concept/language-list/LanguageListTest.php
+++ b/exercises/concept/language-list/LanguageListTest.php
@@ -86,7 +86,7 @@ class LanguageListTest extends TestCase
     /**
      * @task_id 4
      */
-    #[TestDox('when pushing, original is unchanged')]
+    #[TestDox('when removing, original is unchanged')]
     public function testPruningDoesNotMutate()
     {
         $language_list = language_list('c', 'cpp', 'php');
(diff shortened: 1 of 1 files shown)


exercises/concept/language-list/LanguageListTest.php
```diff
@@ -86,7 +86,7 @@ public function testCompleteLanguageList()
     /**
      * @task_id 4
      */
-    #[TestDox('when pushing, original is unchanged')]
+    #[TestDox('when removing, original is unchanged')]
```
mk-mxp: It's better to use the actual function verb 'prune' here, as the list is emptied out completely and not only some things are removed:

```suggestion
    #[TestDox('when pruning, original is unchanged')]
```

mk-mxp (changes requested):
Thanks for pointing this out!

exercises/concept/language-list/LanguageListTest.php
resu-xuniL: I have hesitated ... (I am not English native).
Thanks for pointing this.

mk-mxp: No worries, I'm not a native English speaker, too.

mk-mxp (approved):

PR #999, opened 2026-06-05
Add CI job to ensure practice exercise order

mk-mxp:
Establish a reproducible ordering of exercises, see [forum thread](https://forum.exercism.org/t/sorting-practice-exercises/56068) for details.

I chose sorting by `difficulty` directly, because that allows to use all 9 levels. And lower cased name did not sound bad to me.

Now someone would need to adjust difficulties to match some real difficulty :smiley:

### DIFF
--- a/.github/workflows/order-of-practice-exercises.yml
+++ b/.github/workflows/order-of-practice-exercises.yml
@@ -1,0 +1,16 @@
+name: Ensure practice exercise order
+
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+jobs:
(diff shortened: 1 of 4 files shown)


homersimpsons (approved):
Thank you for the thread link. I have no opinion on sorting here, I'm not even sure why this sorting is used on the site versus sorting them consistently across tracks.

mk-mxp:
If I remember correctly, the argument for track specific order was "Exercises might have different diffculties per track". As people like to solve exercises "in order", I hope they now solve more easy ones before attempting the hard ones.

PR #1003, opened 2026-06-09
Add `Swift scheduling` exercise

resu-xuniL:
Another exercice for validation.

### DIFF
--- a/config.json
+++ b/config.json
@@ -490,6 +490,14 @@
         ]
       },
       {
+        "slug": "swift-scheduling",
+        "name": "Swift Scheduling",
+        "uuid": "d080d214-1d41-4347-a494-1e05db9f614a",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1
+      },
+      {
         "slug": "transpose",
         "name": "Transpose",
         "uuid": "cb413881-99b6-493f-bea2-89563dc173a6",
--- a/config.json
+++ b/config.json
@@ -490,14 +490,6 @@
         ]
       },
       {
-        "slug": "swift-scheduling",
-        "name": "Swift Scheduling",
-        "uuid": "d080d214-1d41-4347-a494-1e05db9f614a",
-        "practices": [],
-        "prerequisites": [],
-        "difficulty": 1
-      },
-      {
         "slug": "transpose",
         "name": "Transpose",
         "uuid": "cb413881-99b6-493f-bea2-89563dc173a6",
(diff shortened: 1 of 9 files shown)


exercises/practice/swift-scheduling/.meta/example.php
```diff
@@ -0,0 +1,104 @@
...
+class SwiftScheduling
+{
+    private const QUARTER = [
+        1 => 4,
+        2 => 7,
+        3 => 10,
+        4 => 13
+    ];
+
+    public function deliveryDate(DateTime $meetingStart, string $description): ?DateTime
```
mk-mxp: This must match the students interface (must not widen the return value type).

```suggestion
    public function deliveryDate(DateTime $meetingStart, string $description): DateTime
```

exercises/practice/swift-scheduling/.meta/example.php
```diff
@@ -0,0 +1,104 @@
...
+                return $this->convertAsap($meetingStart);
+            case 'EOW':
+                return $this->convertEow($meetingStart);
+            default:
+                if (str_ends_with($description, "M")) {
+                    return $this->convertVariableM($meetingStart, (int) substr($description, 0, -1));
+                } elseif (str_starts_with($description, "Q")) {
+                    return $this->convertVariableQ($meetingStart, (int) substr($description, 1,));
+                } else {
+                    return null;
```
mk-mxp: I think that's the least work to replace the `null` return value:

```suggestion
                    throw InvalidArgumentException('Unexpected $description');
```

config.json
```diff
@@ -489,6 +489,14 @@
           "floating_point_numbers"
         ]
       },
+      {
+        "slug": "swift-scheduling",
+        "name": "Swift Scheduling",
+        "uuid": "d080d214-1d41-4347-a494-1e05db9f614a",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1
```
mk-mxp: Looking at your `example.php` code, I'd give this a `5` difficulty. It's quite big and a fair bit of logic to implement.

mk-mxp (changes requested):
Thanks a lot for your contribution! I'm not totally pleased with the class interface, as it makes the problem look like a functional problem (no constructor args, one method to call). But we go with it when you do not want to invest more time / effort.

The introduction sounds more like "one meeting start given, many delivery dates derived" to me. Based on that, I'd suggest changing the implementation to this interface:

```php
class SwiftScheduling
{
    public function __construct(DateTime $meetingStart)
    {
        throw new \BadMethodCallException(sprintf('Implement the %s method', __FUNCTION__));
    }

    public function deliveryDate(string $description): DateTime
    {
        throw new \BadMethodCallException(sprintf('Implement the %s method', __FUNCTION__));
    }
}
```

You can use [constructor promotion](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion) to ease implementation.

resu-xuniL:
I hope I correctly understood your proposal :)

mk-mxp (approved):
Yes, that's what I had in mind. Thanks for your effort!

resu-xuniL:
You're welcome! thanks to you ...

Issue #994, opened 2026-06-02
Sync robot-name


mk-mxp:
- [x] Run `bin/configlet sync -u -e robot-name --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
- [x] Drop strict types comments from test file and example code (these are useless)
  - **Keep the comment in the students stub file**. And the directive itself shall be in all PHP files, too.
- [x] Add and sync test meta data to tests (`uuid` in DocBlocks / `#TestDox()`)
- [x] Add exercise to `bin/auto-sync.txt` to keep it in sync (keep the file alphabetically sorted)
- [x] Decide on adding / adjusting / ordering test cases to [match current problem specs](https://github.com/exercism/problem-specifications/blob/main/exercises/robot-name/canonical-data.json)
- [x] If you wish, add your GitHub handle to the list of contributors in `exercises/practice/robot-name/.meta/config.json`

Do not redesign the student's interface or add test cases that would invalidate existing community solutions. These are extra tasks, which should be discussed in advance.

Hazeolation:
Hello! I'd like to work on this if it's fine

mk-mxp:
You are welcome. Please remember that I will be on vacation from Friday on and it might take a while until I can review.

mk-mxp:
@Hazeolation Are you stuck? Can I help to take this forward? Have your plans changed and this issue is no longer interesting to you?

Hazeolation:
Hi, sorry, I'm in exam phase right now so I would continue when I'm done (July 27th)

mk-mxp:
@Hazeolation Thanks for your feedback! I'll leave it for you, then. We are not in hurry... Good luck for the exams!

resu-xuniL:
@mk-mxp: can you re-open this ? as I didn't sync this exercise in #1031 (for it was assigned to Hazeolation):  I just inlined instantiations.

mk-mxp:
@resu-xuniL Not sure, what you need this issue for. The PR #1005 was merged a month ago, that's it for this issue. I only had forgotten to close this issue. You can do your changes within your issue, with nobody being in your way.

resu-xuniL:
> [@resu-xuniL](https://github.com/resu-xuniL) Not sure, what you need this issue for. The PR [#1005](https://github.com/exercism/php/pull/1005) was merged a month ago, that's it for this issue. I only had forgotten to close this issue. You can do your changes within your issue, with nobody being in your way.

[...]

PR #1005, opened 2026-06-10
#994 Sync robot-name

Hazeolation:
* Synced robot-name metadata and docs
* Added GitHub handle to config.json
* Added robot-name to auto-sync.txt
* Removed strict type comments from `example.php` and `RobotNameTest.php`

Ref: #994

### DIFF
--- a/.github/workflows/configlet-generate.yml
+++ b/.github/workflows/configlet-generate.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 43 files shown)


mk-mxp (approved):
That was unexpectedly easy 😄 with no canonical data. Thanks for taking it!

You seem to have git and GitHub workflow trouble. You should re-sync your GitHub fork and local git repository before doing another issue and use git branches to work on. If you want help with that, open a forum thread or send me a message there. In the forum I can use screenshots. Describing all of the steps here in text-only is very hard...

PR #1008, opened 2026-06-13
Sync triangle

Narkunan:
Fix issue [https://github.com/exercism/php/issues/984](https://github.com/exercism/php/issues/984)


Narkunan:
Hi @mk-mxp I hope You had great with your vacation. I know this build will fail but i want to raise a PR Because I want to discuss some test cases.

1. 8d71e185-2bd7-4841-b7e1-71689a5491d8 – This test case satisfies both the **equilateral** and **isosceles** conditions. How should we handle this case?
2. c6585b7d-a8c0-4ad8-8a34-e21d36f7ad87 – This test case satisfies the **scalene** condition, but it is expected to throw an exception if it is equilateral.
3. 33eb6f87-0498-4ccf-9573-7f8c3ce92b7b – Same as test case 2.
4. 2510001f-b44d-4d18-9872-2303e7977dc1 – This test case satisfies the **equilateral** triangle condition, but it is expected to be treated as **scalene.
5. c6e15a92-90d9-4fb3-90a2-eef64f8d3e1e – This test case is expected to be **scalene** and throw an exception, but it also matches the **isosceles condition.
6. 3da23a91-a166-419a-9abf-baf4868fd985 – Same as test case 5.
7. b6a75d98-1fef-4c42-8e9a-9db854ba0a4d– Same as test case 5.

mk-mxp:
Thanks for taking this. It looked much more easy than it actually is. I'm sorry if this is out of your capabilities (time, effort, difficulty).

Just in case I haven't said that before: There is a common repository that is used as a data source for the test cases: [canonical data for triangle](https://github.com/exercism/problem-specifications/blob/main/exercises/triangle/canonical-data.json).

The problem you are facing is, that the current solution interface does not match the problem anymore. We should not ask for one `kind()`, but redesign to ask for several properties of the `Triangle` object: `isEquilateral()`, `isIsosceles()`, `isScalene()`.

So, first rewrite the students stub file to contain this class interface:

```php
class Triangle
{
    public function __construct(int $a, int $b, int $c)
    {
        throw new \BadMethodCallException(sprintf('Implement the %s method', __FUNCTION__));
    }

    public function isEquilateral(): bool
    {
        throw new \BadMethodCallException(sprintf('Implement the %s method', __FUNCTION__));
    }

    public function isIsosceles(): bool
    {
        throw new \BadMethodCallException(sprintf('Implement the %s method', __FUNCTION__));
    }

    public function isScalene(): bool
    {
        throw new \BadMethodCallException(sprintf('Implement the %s method', __FUNCTION__));
    }
}
```

Then you have to rewrite all tests and the `example.php` to use this interface. Exceptions shouldn't be required anymore.

Are you willing to do so? You can ask for help at any time. But it's also OK to say "no", as this is not an easy synchronisation anymore.

mk-mxp:
To keep this an easy exercise, we should not add the `floating-point` scenario tests. Mark the 3 test cases as skipped in `exercises/practice/triangle/.meta/tests.toml` like this:

```toml
[3022f537-b8e5-4cc1-8f12-fd775827a00c]
description = "equilateral triangle -> sides may be floats"
include = false
comment = "comparing floats would make the exercise too difficult"
```

Narkunan:
@mk-mxp Thanks! I cannot fix it for another 5 days. I am little busy with my work and personal commitment. I will try to do as soon as possible

Narkunan:
@mk-mxp Updated based on suggestion. pleas review it when you have a time. Thanks!

mk-mxp (approved):
Thanks a lot! That looks great.

PR #1009, opened 2026-06-30
Increase tolerance for `BafflingBirthdaysTest.php`

resu-xuniL:
Damned randomness !

### DIFF
--- a/exercises/practice/baffling-birthdays/BafflingBirthdaysTest.php
+++ b/exercises/practice/baffling-birthdays/BafflingBirthdaysTest.php
@@ -198,7 +198,7 @@ class BafflingBirthdaysTest extends TestCase
         $this->assertEqualsWithDelta(
             0.0,
             $birthdays->estimatedProbabilityOfSharedBirthday(1),
-            1
+            2
         );
     }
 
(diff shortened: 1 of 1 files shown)


exercises/practice/baffling-birthdays/BafflingBirthdaysTest.php
mk-mxp: Shouldn't this be 2, too?

mk-mxp: Why not 2 here?

mk-mxp (changes requested):

exercises/practice/baffling-birthdays/BafflingBirthdaysTest.php
resu-xuniL: I never needed to increase this tolerance value : always got 98-100 %
Should I nevertheless increase it for safety?

resu-xuniL: same here, always got 0%. That's why I started to assert without delta at first.
(-> there can't be 2 same dates as only 1 is generated)

mk-mxp: It should be wide enough to accept wildly different ways to calculate the probability. It's "not much data", so I think a bigger tolerance would be better in both places.

mk-mxp (approved):
Thank you! I hope it doesn't come back again now...

resu-xuniL:
So do I !
Thanks to you.

PR #1011, opened 2026-07-01
Add `square-root` exercise

resu-xuniL:
Another exercice for validation.

### DIFF
--- a/config.json
+++ b/config.json
@@ -490,6 +490,14 @@
         ]
       },
       {
+        "slug": "square-root",
+        "name": "Square Root",
+        "uuid": "d75c009b-ad65-455f-a824-ed61c8797cc4",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1
+      },
+      {
         "slug": "transpose",
         "name": "Transpose",
         "uuid": "cb413881-99b6-493f-bea2-89563dc173a6",
--- a/exercises/practice/square-root/.meta/example.php
+++ b/exercises/practice/square-root/.meta/example.php
@@ -1,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+function squareRoot(int $number): int
+{
+    $sqrt = $number;
+
+    while (pow($sqrt, 2) > $number) {
+        $sqrt--;
+    }
+
+    return $sqrt;
+}
--- a/exercises/practice/square-root/SquareRoot.php
+++ b/exercises/practice/square-root/SquareRoot.php
@@ -1,0 +1,30 @@
+<?php
+
+/*
+ * By adding type hints and enabling strict type checking, code can become
+ * easier to read, self-documenting and reduce the number of potential bugs.
+ * By default, type declarations are non-strict, which means they will attempt
+ * to change the original type to match the type specified by the
+ * type-declaration.
+ *
+ * In other words, if you pass a string to a function requiring a float,
+ * it will attempt to convert the string value to a float.
+ *
+ * To enable strict mode, a single declare directive must be placed at the top
+ * of the file.
+ * This means that the strictness of typing is configured on a per-file basis.
+ * This directive not only affects the type declarations of parameters, but also
+ * a function's return type.
+ *
+ * For more info review the Concept on strict type checking in the PHP track
+ * <link>.
+ *
+ * To disable strict typing, comment out the directive below.
+ */
+
+declare(strict_types=1);
+
+function squareRoot(int $number): int
+{
+    throw new \BadMethodCallException(sprintf('Implement the %s function', __FUNCTION__));
+}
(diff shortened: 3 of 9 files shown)


exercises/practice/square-root/.meta/example.php
```diff
@@ -0,0 +1,14 @@
...
+
+function squareRoot(int $number): int
+{
+    $sqrt = $number;
+
+    while (pow($sqrt, 2) > $number) {
+        $sqrt--;
+    }
+
+    return $sqrt;
```
mk-mxp: I doubt this is more efficient than using a math library 😺 Anyways, it works...

exercises/practice/square-root/SquareRoot.php
```diff
@@ -24,19 +24,7 @@
 
 declare(strict_types=1);
 
-class GameOfLife
+function squareRoot(int $number): int
```
mk-mxp, line 27: Yeah! Functional interface for a functional problem.

exercises/practice/square-root/SquareRootTest.php
```diff
@@ -0,0 +1,68 @@
...
+use PHPUnit\Framework\TestCase;
+use PHPUnit\Framework\Attributes\TestDox;
+
+class SquareRootTest extends TestCase
+{
+    public static function setUpBeforeClass(): void
+    {
+        require_once 'SquareRoot.php';
+    }
+
```
mk-mxp, line 14: Maybe the first test should be a simple test against using PHP functions `*sqrt()`? Something like:

```php
#[TestDox('Does not use or mention PHP square root functions')]
public function testDoesNotUseOrMentionPhpSquareRootFunctions(): void
{
    $code = file_get_contents('SquareRoot.php');

    $this->assertStringNotContainsString('sqrt', $code);
}
```

Nothing too fancy, but against very simple use cases. What do you think?

config.json
```diff
@@ -489,6 +489,14 @@
           "floating_point_numbers"
         ]
       },
+      {
+        "slug": "square-root",
+        "name": "Square Root",
+        "uuid": "d75c009b-ad65-455f-a824-ed61c8797cc4",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1
```
mk-mxp: Math is not for everybody, maybe `2`?

mk-mxp (changes requested):
Looks good to me. If you want me to keep it as it is, I'll merge it.

exercises/practice/square-root/SquareRootTest.php
resu-xuniL: That's a good idea.

resu-xuniL: I add a `(` to match function and not `sqrt` word as one can use this naming for the solution variable (as I did)

exercises/practice/square-root/SquareRootTest.php
```diff
@@ -0,0 +1,76 @@
...
+    #[TestDox('does not use or mention PHP square root functions')]
+    public function testDoesNotUseOrMentionPhpSquareRootFunctions(): void
+    {
+        $code = file_get_contents(__DIR__ . '/SquareRoot.php');
+
+        $this->assertStringNotContainsString('sqrt(', $code);
```
mk-mxp: I intentionally omitted the `(`, because that only catches `sqrt(` and not `sqrt (`. Which is a very simple thing to do. So I'd rather rename the variable in `example.php` to `$result`...

Maybe an additional message is more polite to the students:

```suggestion
        $this->assertStringNotContainsString('sqrt', $code, 'Please do not use "sqrt" anywhere in your code!');
```

mk-mxp (changes requested):

exercises/practice/square-root/SquareRootTest.php
resu-xuniL: Oh I see!

config.json
```diff
@@ -577,6 +577,14 @@
           "math"
         ]
       },
+      {
+        "slug": "crypto-square",
+        "name": "Crypto Square",
+        "uuid": "14aa815d-8073-4a71-841e-9bfdba4b0b1a",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 2
```
mk-mxp: Sorry, I didn't see this before...I think you accidentaly changed the wrong difficulty here. We are on `square-root`, not `crypto-square`.

mk-mxp (changes requested):

config.json
resu-xuniL: You're right! sorry
I wondered why but couldn't find the slug then didn't investigate further ... my bad!

mk-mxp (approved):
Thanks a lot for your work!

resu-xuniL:
My pleasure. Thanks to you.

PR #1016, opened 2026-07-11
Sync `transpose` (issue #985)

### LINKED ISSUE #985: Sync transpose
- [ ] Run `bin/configlet sync -u -e transpose --yes --docs --filepaths --metadata --tests include` (updates the Markdown files and maybe `tests.toml`)
  - docs: instructions unsynced: transpose
  - metadata: unsynced: transpose
- [ ] Drop strict types comments from test file and example code (these are useless)
  - **Keep the comment in the students stub file**. And the directive itself shall be in all PHP files, too.
- [ ] Add and sync test meta data to tests (`uuid` in DocBlocks / `#TestDox()`)
- [ ] Add exercise to `bin/auto-sync.txt` to keep it in sync (keep the file alphabetically sorted)
- [ ] Decide on adding / adjusting / ordering test cases to match current problem specs
  - transpose: missing 1 test case
- [ ] If you wish, add your GitHub handle to the list of contributors in `exercises/practice/transpose/.meta/config.json`

resu-xuniL:
- [x] Synced `Transpose` metadata and docs
- [x] Removed strict types comments from `example.php` and `TransposeTest.php`
- [x] Added `transpose` to `auto-sync.txt`
- [x] Added UUID and TestDox to tests
- [x] Fixed `example.php` to pass new test unit

Had to modify `example.php` to pass new test cases from [problem-specifications](https://github.com/exercism/problem-specifications/blob/main/exercises/transpose/canonical-data.json)

Closes #985


mk-mxp (changes requested):
Let's make this a difficulty `2` exercise. The padding is a real challenge in itself, I think.

mk-mxp (approved):
This also may break some solutions, but the requirements were there always. So, I think that's OK.

Thanks a lot for your work!

resu-xuniL:
My pleasure! thanks.

resu-xuniL:
I think this PR will break the new 'sorting use' rule

PR #1026, opened 2026-07-12
Bump workflows to Ubuntu 26.04

IsaacG:


mk-mxp:
Please be aware of https://github.com/shivammathur/setup-php/issues/1083 : Ubuntu 26.04 currently is not supported.

IsaacG:
I was unaware of that. Thanks for bringing it up!

The checks pass, though, so it seems to be working, which confuses me a bit here.

mk-mxp:
It's a tricky thing: You currently cannot install PHP8.5 (which we will have to support soon) to Ubuntu 26.04 from Ondrej Sury's repos (which are what shivamatur uses) as well as many PHP packages for PHP < 8.5 (which we do not use). So we are lucky that it happens to work right now, but it will block us when upgrading. I have that problem at work, too ... they are working on it.

Anyways, I'll leave this opened. There will be a time to switch to Ubuntu 26.04 for us. And to get PHP 8.5 supported.

IsaacG:
Sounds good. Thank you!

mk-mxp:
The workflow owner confirmed that it works now. Thank you!

mk-mxp (approved):

PR #1033, opened 2026-07-19
Add `swift scheduling`  test  + update `perfect-numbers` docs

resu-xuniL:
following [problem-specifications](https://github.com/exercism/problem-specifications/blob/main/exercises/swift-scheduling/canonical-data.json)

`swift scheduling` got a new test + adjust `example.php`

(and update `perfect-numbers` docs)


mk-mxp (approved):
Thanks for helping! Such a small new test case and so much to correct...

resu-xuniL:
You're welcome!
I must admit that only one change would have been necessary for the new test (the `round()` replacement with `intdiv()`). But I have seized the occasion to remove useless code :)

PR #1039, opened 2026-07-22
Add `intergalactic transmission` exercise

resu-xuniL:
Hello!
Here is another exercise for validation:

- [x] Generated exercice  with configlet create
- [x] Coded test and example
- [x] Added exercise to `auto-sync.txt`
- [x] Set the difficulty rank to 3

Is dificulty rank 3 OK?

### DIFF
--- a/bin/auto-sync.txt
+++ b/bin/auto-sync.txt
@@ -37,6 +37,7 @@ hamming
 hello-world
 high-scores
 house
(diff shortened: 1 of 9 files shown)


mk-mxp (approved):
This looks great! We will see, if the longer array rows will show up nicely in the online editor. They don't look tooooo long.

PR #1041, opened 2026-08-02
Ease contribution with Devcontainer

mk-mxp:
Provide a Devcontainer setup for contributors. This provides all required PHP modules / extensions.


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+    && apt-get -y install --no-install-recommends jq \
+    && apt-get clean -y && rm -rf /var/lib/apt/lists/*
+
+# ds is pinned to v1, since v2 (the only version building against PHP 8.4 via
+# PECL/docker-php-ext-install) dropped Vector/Deque/Stack/Queue/PriorityQueue.
+RUN curl -L -o install-php-extensions \
+    https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions \
+    && chmod +x install-php-extensions \
+    && ./install-php-extensions gmp ds-^1@stable intl \
+    && rm install-php-extensions
```
homersimpsons, line 13: Should we instead upgrade the test runner and everything to V2 ? Or would you prefer doing this in the php 8.5 upgrade ?

Note that even if they were removed, new alternatives were provided https://github.com/php-ds/ext-ds/releases/tag/v2.0.0

README.md
```diff
@@ -5,16 +5,26 @@ Exercism exercises in PHP
...
 
 The following system dependencies are required:
 
 - `composer`, as recommended in the [PHP track installation docs][exercism-track-installation-composer].
 - [`bash` shell][gnu-bash].
-- PHP V8.4+ CLI.
+- PHP V8.4+ CLI, with the following extensions:
+  - `ds` (V1.x)
+  - `intl`
+  - Default modules: `Core`, `ctype`, `date`, `dom`, `fileinfo`, `filter`, `hash`, `iconv`, `json`, `libxml`, `mbstring`, `pcre`, `random`, `Reflection`, `SimpleXML`, `sodium`, `SPL`, `standard`, `tokenizer`, `xml`, `xmlreader`, `xmlwriter`, `zlib`
```
homersimpsons, line 22: Shouldn't we just mention `ext-*` requirements? It does not look relevant to mention "Core", "ctype", "date"...

homersimpsons (approved):

.devcontainer/Dockerfile
mk-mxp: `ds` is not a requirement the track has or I introduced. It was "requested" by neenjaw (the maintainer before I came in) as a useful computer-sciency data type extension used in community solutions.

I did not upgrade it because of the breaking interface changes. I'd rather "discuss" (if there is anyone discussing that with me anyways) to drop it completely with the PHP 8.5 upgrade - it's not there for fluency in PHP, but for academic knowledge and production grade programming.

README.md
mk-mxp: It is the same list we added to the test-runner. It is the same list, that shall be added to the PHP track docs. A beginner who wants to start learning can use `php -m` to check the installed modules using these names. With composers `ext-*` notation they have to translate the names to composers style, have composer installed at all and filter away many more names from the composer output than from `php -m`.

To us, `ctype` et al. do not seem relevant. In PHP documentation, the functions and classes **are** organized by them. So I can use [the `ctype` functions](https://www.php.net/manual/en/book.ctype.php), study them, try them out in my solutions etc. It's a common question: What functions could I use at all? What's my "search space"?

.devcontainer/Dockerfile
mk-mxp: See [forum](https://forum.exercism.org/t/remove-php-extension-ds/69730) for follow-up.

PR #1042, opened 2026-08-02
AI support for contributors

mk-mxp:
I am not sure, if we should add these. If we decide not to help contributors with using AI for contributions, I'm OK with that.

A main instructions file for the track's day-to-day work, and one for the contribution tooling. It saves a few tokens for the future.

And a minor fix for the PHP version range documented in README, nothing functional.


CLAUDE.md
homersimpsons, line 1: While I personally find a lot of issues with LLM outputs quality (and their societal / environmental impacts). I'm not completely against introducing this. BUT:
1. I would prefer AGENTS.md (over CLAUDE.md) to avoid favouring (in terms of visibility) any AI provider. I know that `@iHiD ` uses Claude though.
2. I would prefer we use no files if the informations are relevant for humans too. In which case we should seek concise writing and use classical README.md / CONTRIBUTING.md files
3. Similarly I would not gitignore the local LLM files, in my opinion the developers should ignore them globally instead. Though if we face a lot of good quality merge requests with local files included I will reconsider adding it.

mk-mxp: I'm not a fan of LLMs, too. I currently learn to use them (as we all have to), and a project without agent supporting files burns a lot more tokens (energy) than a project with them. Contributors use them anyways already, so let's save some tokens here.

To 1: I know, that Claude does not read AGENTS.md on purpose. That's a shame, but they are current market leaders (people urged to use Copilot for free on GitHub is no market share). I do not have another coding agent at hand to test support (cursor et al.). So, I can only reliably provide Claude information and supporting files.

To 2: Agent files are auto-generated and maintained by the agents. They read the README and code, and write their prose to safe tokens for the future. I do not intend to remove / reduce human first information anywhere.

To 3: gitignoring local agent files is common practice like `.env.local.*` already are. Especially with the many CDEs (like GitHub codespaces) global settings are unreliable. They also may contain security information (like tokens, keys for MCPs), as they are personal files.

mk-mxp: @homersimpsons Can you tell me what I can do to take this conversation forward? I have no other AI / LLM tools than Claude to verify anything regarding AGENTS.md or multi-LLM support.

I could add LLM instructions so they see READMEs as first class sources of truth, but that does Claude do anyways. We can remove the personal (LLM) settings from the .gitignore, but then we should also remove `.idea` as well as the other PHP track unrelated spam sources like `DS_store`.

What do you think?

PR #1048, opened 2026-08-25
Add `split second stopwatch` exercise

resu-xuniL:
Hello!
Here is another exercise for validation:

- [x] Generated exercice  with configlet `create`
- [x] Coded test and example
- [x] Set the difficulty rank to 3

Is dificulty rank 3 OK?


exercises/practice/split-second-stopwatch/SplitSecondStopwatch.php
```diff
@@ -0,0 +1,68 @@
...
+ * To disable strict typing, comment out the directive below.
+ */
+
+declare(strict_types=1);
+
+class SplitSecondStopwatch
+{
+    public function __construct()
+    {
+        throw new \BadMethodCallException(sprintf('Implement the SplitSecondStopwatch %s method', __FUNCTION__));
```
mk-mxp: Use [`__METHOD__`](https://www.php.net/manual/en/language.constants.magic.php#constant.method) here:


```suggestion
        throw new \BadMethodCallException(sprintf('Implement the SplitSecondStopwatch %s method', __METHOD__));
```

exercises/practice/split-second-stopwatch/SplitSecondStopwatch.php
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * For more info review the Concept on strict type checking in the PHP track
+ * <link>.
+ *
+ * To disable strict typing, comment out the directive below.
+ */
+
+declare(strict_types=1);
+
+class SplitSecondStopwatch
```
mk-mxp, line 27: Please add a comment about visibility (`private(set)`) and that properties must be defined by the student.

mk-mxp (changes requested):
This looks great! If you don't want to raise the difficulty a bit as described below, we can merge it with `__METHOD__` and the property comment only.

To raise the difficulty to `5` you could remove the method stubs from the student file's class body and place a comment there, that all methods need to be defined by the students. This surprises students often and we do this not very often, yet.

mk-mxp (approved):
Thanks for the great exercise!

resu-xuniL:
> Thanks for the great exercise!

You're welcome!

PR #1053, opened 2026-09-21
Add `complex-numbers` exercise

resu-xuniL:
Hello!
Here is `complex-numbers` exercise for validation:

- [x] Generate exercice files with `configlet create`
- [x] Create `test` and `example` files
- [x] Set the difficulty rank to `6`
- [x] Add GitHub handle to `config.json`

Is difficulty rank 6 OK?

I'm not really pleased with the 8 last tests:
I had to create the `addR` `subR` `mulR` `divR` methods to manage a `real` number.

I looked how the javascript track managed it: they are passing the `real number` as a `complex number` (so they can use only 4 methods, one for each arithmetic operation)
```
    xtest('Subtract real number from complex number', () => {
      const expected = new ComplexNumber(1, 7);
      const actual = new ComplexNumber(5, 7).sub(new ComplexNumber(4, 0));

      expect(actual.real).toBeCloseTo(expected.real);
      expect(actual.imag).toBeCloseTo(expected.imag);
    });

    xtest('Subtract complex number from real number', () => {
      const expected = new ComplexNumber(-1, -7);
      const actual = new ComplexNumber(4, 0).sub(new ComplexNumber(5, 7));

      expect(actual.real).toBeCloseTo(expected.real);
      expect(actual.imag).toBeCloseTo(expected.imag);
    });

```

Should we do the same thing? i.e. get rid of the 4 methods I created (the ones finishing with a capital `R`) and keeping only the 4 methods `add` `sub` `mul` `div` such as in the Javascript track ?

This would gives:
```
    public function testOperationsBetweenRealNumbersAndComplexNumbersSubtractRealNumberFromComplexNumber(): void
    {
        $realNumber = new ComplexNumbers(4); // --> instead of: $realNumber = 4;
        $ComplexNumber = new ComplexNumbers(5, 7);
        $expected = new ComplexNumbers(1, 7);

        $this->assertEquals($expected, $ComplexNumber->sub($realNumber));
        // --> instead of: $this->assertEquals($expected, $ComplexNumber->subR($ComplexNumber, $realNumber));
    }


    public function testOperationsBetweenRealNumbersAndComplexNumbersSubtractComplexNumberFromRealNumber(): void
    {
        $realNumber = new ComplexNumbers(4); // --> instead of: $realNumber = 4;
        $ComplexNumber = new ComplexNumbers(5, 7);
        $expected = new ComplexNumbers(-1, -7);

        $this->assertEquals($expected, $realNumber->sub($ComplexNumber)); 
        // --> instead of: $this->assertEquals($expected, $ComplexNumber->subR($realNumber, $ComplexNumber));
    }
```
What do you think ?


IsaacG:
+cc @exercism/php

mk-mxp:
@IsaacG It's on the radar, but time is scarce...
@resu-xuniL I'll take a look these days, as soon as time is available.

homersimpsons:
About CI:
- Test passes for this exercise: https://github.com/exercism/php/actions/runs/35911404438/job/107351853716?pr=1053#step:6:268
- macOS failure are not related to the changes here
- i18n is failing because it is expecting a `metadata/php.json` in `exercism/i18n`, I do not really know what is expected here

IsaacG:
Re: i18n, see https://github.com/exercism/problem-specifications/issues/2686

TL;DR: finalize the PR and get approval first. Once it's ready to merge, a label is used to trigger the translation, which should do a bunch of i18n work and clear the CI.

exercises/practice/complex-numbers/.meta/example.php
```diff
@@ -0,0 +1,104 @@
...
+
+    public function addR(int|ComplexNumbers $num1, int|ComplexNumbers $num2): ComplexNumbers
+    {
+        is_int($num1) ? $num1 = new ComplexNumbers($num1) : $num2 = new ComplexNumbers($num2);
+
+        return new ComplexNumbers(
+            $num1->real + $num2->real,
+            $num1->imaginary + $num2->imaginary
+        );
+    }
```
homersimpsons: In my opinion `addR` should not exists, and can be purely deleted in favour of `add`.

In fact:
1. The canonical data does not mention another `property`
2. Figuring out that one should add `addR` feels wrong to me
3. Looking at others track they are not bothering the student with accepting `int`egers: [javascript](https://github.com/exercism/javascript/blob/main/exercises/practice/complex-numbers/complex-numbers.spec.js), [java](https://github.com/exercism/java/blob/main/exercises/practice/complex-numbers/src/test/java/ComplexNumberTest.java).

This comment applies for `subR`, `mulR` and `divR`.

homersimpsons (changes requested):
Everything looks good except for the interface we are proposing to the student. Let's align this with other tracks and it is good to go in my opinion.

Thanks @resu-xuniL

exercises/practice/complex-numbers/.meta/example.php
IsaacG: I'm not familiar with PHO, but doesn't this test in the canonical data imply/require `addR`?

https://github.com/exercism/problem-specifications/blob/main/exercises/complex-numbers/canonical-data.json#L391-L397

IsaacG: FWIW the Python `example.py` defines `__radd__()`: https://github.com/exercism/python/blob/main/exercises/practice/complex-numbers/.meta/example.py#L23

homersimpsons: > I'm not familiar with PHO _[PHP],_ but doesn't this test in the canonical data imply/require addR?

I read this test. The representation `"z1": 5,` is in fact different than `"z1": [1, 0],` in other tests. But the mentioned property is `add` also here.

I originally wanted to suggest to remove `addR` and implements `add` so it supports both, then I checked some other tracks and figured out they didn't bother with this.

It is not a good practice to have functions that accept too broad parameters and deals with it using `if`s, it is better to ["push ifs up"](https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html). Here my suggestion was that we would push the `if`s to the test instead of down inside the function.

> FWIW the Python `example.py` defines `__radd__()`:

That's a good "counter-example" but quite in the between of what we had and what java/javascript tracks proposes, in fact in thy python example only `other` has to be considered to be real while the canonical data have both [`Complex+Real` and `Real+Complex`](https://github.com/exercism/problem-specifications/blob/82d94a395a24ae08e4b6a29bb8822c26eaac9ac9/exercises/complex-numbers/canonical-data.json#L379-L398) so our current `addR` would be closer to the specification.

So my order of preference (from highest to lowest):
| Proposed interface                                                                        | Specification           | Interface                            |   |   |
|-------------------------------------------------------------------------------------------|-------------------------|--------------------------------------|---|---|
| `add(ComplexNumbers , ComplexNumbers)`                                                    | -2 (no `int+` / `+int`) | -0                                   |   |   |
| `add(int\|ComplexNumbers, int\|ComplexNumbers)`                                           | -0                      | -2 (2x`int\|ComplexNumbers`)         |   |   |
| `add(ComplexNumbers , ComplexNumbers)` + `addR(ComplexNumbers, int)`                      | -2 (no `int+`, `addR`)  | -1 (`addR`)                          |   |   |
| `add(ComplexNumbers , ComplexNumbers)` + `addR(int\|ComplexNumbers, int\|ComplexNumbers)` | -1 (`addR`)             | -3 (`addR`, 2x`int\|ComplexNumbers`) |   |   |
| `add(ComplexNumbers , ComplexNumbers)` + `addR(ComplexNumbers, int\|ComplexNumbers)`      | -2 (`addR`, no `int+`)  | -2 (`addR`, `int\|ComplexNumbers`)   |   |   |)

IsaacG: > I read this test. The representation `"z1": 5,` is in fact different than `"z1": [1, 0],` in other tests. But the mentioned property is `add` also here.

The canonical data describes the test data, not the implementation details. If the implementation needs `lAdd` and `rAdd` or `add` and `addR` or some other naming, that shouldn't be reflected in the canonical data but should show up in the track tests/implementations. The canonical data makes it pretty clear that the tests expect to be able to do both a `Complex + int` as well as `int + Complex` ... with the implementation details left to the tracks.

